PR #4, opened 2017-09-21
Add websocket to agent-js (subscribed by agent-client-js)

piedup:
Creates a websocket for agent-client-js on ws://localhost:3000 when http server is started. storeHttpClient events are propagated to web socket.

Use `const server = Agent.listen(agent, storeHttpClient)` to start server and websocket.

TODO: Implementation on agent-client-js

### DIFF
--- a/packages/agent-js/README.md
+++ b/packages/agent-js/README.md
@@ -42,19 +42,21 @@ agent.addProcess("my_first_process", actions, storeHttpClient, fossilizerHttpCli
 });
 
 // Creates an HTTP server for the agent with CORS enabled.
-var agentHttpServer = agent.httpServer({ cors: {} });
+var agentHttpServer = Agent.httpServer(agent, { cors: {} });
 
 // Create the Express server.
-var app = express();
-
+const app = express();
 app.disable('x-powered-by');
 
 // Mount agent on the root path of the server.
 app.use('/', agentHttpServer);
 
-// Start the server.
-app.listen(3000, function() {
-  console.log('Listening on :' + this.address().port);
+// Create server by binding app and websocket connection
+const server = Agent.websocketServer(app, storeHttpClient);
+
+// Start the server
+server.listen(3000, () => {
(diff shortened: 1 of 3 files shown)


packages/agent-js/src/listen.js
```diff
@@ -0,0 +1,58 @@
...
+    wss.clients.forEach((client) => {
+      client.send(msgJson);
+    });
+  });
+  wss.on('connection', (ws) => {
+    ws.send('Connected to agent client websocket');
+  });
+
+  // Start the server.
+  server.listen(3000, () => {
```
such: we should probably make it possible to configure the port

such: Actually I think it'd be nice to mirror the httpServer function so that the user has more freedom in the way he wants to arrange things together. I was thinking something like:
```
function websocketServer(httpServer, storeHttpClient) {
  const server = http.createServer(httpServer);
  const wss = new websocket.Server({ server });
 
  // Send messages on storeHttpClient message event
  storeHttpClient.on('message', (msg) => {
    const msgJson = JSON.stringify(msg);
    wss.clients.forEach((client) => {
      client.send(msgJson);
    });
  });
  wss.on('connection', (ws) => {
    ws.send('Connected to agent client websocket');
  });

  return server
}
```

What do you think?
Usage would look like:
```
// Creates an HTTP server and WS server for the agent with CORS enable
var agentHttpServer = agent.httpServer({ cors: {} });
var agentWsServer = websocketServer(agentHttpServer, storeHttpClient)
		
// Create the Express server.		
var app = express();
app.disable('x-powered-by');		
		
// Mount agent on the root path of the server.		
app.use('/', agentWsServer);		
		
// Start the server.		
app.listen(3000, function() {		
  console.log('Listening on :' + this.address().port);		
});
```

piedup:
Rebased the commits and changed the implementation to keep the logic as close as possible to current implementation (see README.md).

PR #14, opened 2017-09-26
agent-js: fix "DISABLED" evidence sent to tmstore

### LINKED ISSUE #5: agent-js: evidence is sent as DISABLED to the store
### Actual behavior

The tmstore transaction contains state: DISABLED
This occurs when no `fossilizerHttpClient` is passed to the agent. In this case, the evidence will (currently) only be edited if the `tmstore` adapter is being used. If we implement some other blockchain store adapter, like hyperledger fabric, the evidence will also be modified after it is sent by the agent.

- Do we want to send an empty evidence to the store ? Or do we want to send `evidence: "QUEUED"` in segment.meta if the store adapter is `tmstore`  and "DISABLED" in other cases ?

simonvadee:
Fixes #5 
Typo in README has been fixed.

### DIFF
--- a/packages/agent-js/src/process.js
+++ b/packages/agent-js/src/process.js
@@ -24,7 +24,6 @@ import generateSecret from './generateSecret';
 import getDefinedFilters from './getDefinedFilters';
 
 const QUEUED = 'QUEUED';
-const DISABLED = 'DISABLED';
 const COMPLETE = 'COMPLETE';
 
 export default class Process {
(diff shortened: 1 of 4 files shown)


simonvadee:
yes it does, since { state: "PENDING" } is already added by tmpop in the evidence field

packages/agent-js/src/process.js
```diff
@@ -200,9 +194,7 @@ export default class Process {
     return this.storeClient
       .getSegment(this.name, linkHash)
       .then(segment => {
-        Object.assign(segment.meta.evidence, evidence, {
-          state: COMPLETE
-        });
+        Object.assign(segment.meta.evidence, Object.assign(evidence, { state: COMPLETE }));
```
such: What does this change?

simonvadee: my bad, I forgot to change it back after my last commit.

Issue #17, opened 2017-09-28
agent-client-js: findSegments using tags does not work


jeremie-stratumn:
### Expected behavior
`process.findSegments({ tags: ['myTag']})` should return segments tagged with `'myTag'`

### Actual behavior
currently does not seem to work properly

### Steps to reproduce the behavior
see test:
https://github.com/stratumn/indigo-js/blob/2997cac0aadd47cabaea46fdc8bc74b6842ee87b/packages/agent-client-js/test/findSegments.js#L33

I believe this is due to https://github.com/stratumn/indigo-js/blob/2997cac0aadd47cabaea46fdc8bc74b6842ee87b/packages/agent-client-js/src/makeQueryString.js#L22
used over http.. Note that the problem is not there when using agent client over an object.

jeremie-stratumn:
This is what I thought. Looking at the url generated by https://github.com/stratumn/indigo-js/blob/c215eba1c869c6a391dca71d3da74e3a2fb52f7d/packages/agent-client-js/src/findSegments.js#L43
it looks like: `http://localhost:3333/first_process/segments?tags=myTag`
When I add 2 tags: `process.findSegments({ tags: ['myTag', 'myTag2']})` it looks like: `http://localhost:3333/first_process/segments?tags=myTag%2BmyTag2`

And on the agent httpServer side this is what is deserialized:
`{ tags: 'myTag' }` and `{ tags: 'myTag+myTag2' }` respectively.

It should an array so:
`{ tags: '[myTag]' }` and `{ tags: '[myTag, myTag2]' }`

PR #18, opened 2017-09-28
agent-client-js: use qs for querystring encoding to handle arrays

jeremie-stratumn:
https://github.com/stratumn/indigo-js/issues/17

use node library qs to encode query strings which is also used by express extended query parser to decode query strings. See: http://expressjs.com/en/api.html

### DIFF
--- a/packages/agent-client-js/package.json
+++ b/packages/agent-client-js/package.json
@@ -61,6 +61,7 @@
   },
   "dependencies": {
     "httpplease": "^0.16.4",
+    "qs": "^6.5.1",
     "setimmediate": "^1.0.5",
     "xmlhttprequest": "~1.7.0"
   }
--- a/packages/agent-client-js/src/makeQueryString.js
+++ b/packages/agent-client-js/src/makeQueryString.js
@@ -21,7 +21,11 @@ import { stringify } from 'qs';
  * @returns {string} a query string
  */
 export default function makeQueryString(obj) {
-  const query = stringify(obj);
+  // use brackets format for compatibility with GO:
+  // https://github.com/google/go-querystring
+  // see also conversation in:
+  // https://github.com/stratumn/indigo-js/pull/18
+  const query = stringify(obj, { arrayFormat: 'brackets' });
   if (query.length) {
     return `?${query}`;
   }
--- a/packages/agent-client-js/src/makeQueryString.js
+++ b/packages/agent-client-js/src/makeQueryString.js
@@ -13,6 +13,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 */
+import { stringify } from 'qs';
 
 /**
  * Makes a query string.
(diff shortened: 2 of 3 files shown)


stephan83:
So what does a serialized array look like?

jeremie-stratumn:
qs will encode this object ` { tags: [ 'myTag', 'myTag2' ] }` like this:
`tags%5B0%5D=myTag&tags%5B1%5D=myTag2`
which is really:
`tags[0]=myTag&tags[1]=myTag2`

I believe this is compatible with https://github.com/google/go-querystring

stephan83:
See: https://github.com/stratumn/sdk/issues/186

stephan83:
From the go-querystring docs:

> Slice and Array values default to encoding as multiple URL values of the same name. Including the "comma" option signals that the field should be encoded as a single comma-delimited value. Including the "space" option similarly encodes the value as a single space-delimited string. Including the "semicolon" option will encode the value as a semicolon-delimited string. Including the "brackets" option signals that the multiple URL values should have "[]" appended to the value name. "numbered" will append a number to the end of each incidence of the value name, example: name0=value0&name1=value1, etc.

So I believe it can parse this format:

```
tags[]=myTag&tags[]=myTag2
```

I'm not sure it can parse the format you described (with numbers in the brackets).

jeremie-stratumn:
qs can do that too:

```
qs.stringify({ a: ['b', 'c', 'd'] }, { indices: false });
// 'a=b&a=c&a=d'
```

see https://github.com/ljharb/qs

stephan83:
Yes, it can also do brackets without numbers, let's  just pick one compatible with both :)

jeremie-stratumn:
yes that one then! I'll make the change

stephan83:
Yes I think brackets are better, hopefully that way `qs` will parse it as an array even if there is only one value.

jeremie-stratumn:
yes that should work and this test should check that: https://github.com/stratumn/indigo-js/pull/18/files#diff-f0dcd32587ea7ba6c47eb39a7476b8cfR58

jeremie-stratumn:
sorry this is the right example:

```
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'brackets' })
// 'a[]=b&a[]=c'
```

you were right that the previous version was not capable to decode an array of size 1... :)

jeremie-stratumn:
running my test I now see:
tags%5B%5D=myTag&tags%5B%5D=myTag2
tags%5B%5D=myTag

which is what we wanted

stephan83:
I think go-querystring might expect the brackets not to be url-encoded, we'll see.

stephan83:
Well that Go library only encodes values, it doesn't parse querystrings :(

stephan83:
Should be good on the Go side, but we also need to rebuild all the libraries that use agent-client-js.

PR #20, opened 2017-09-29
agent-js: use brackets for query arrays

stephan83:
`storeHttpClient` will now use brackets when making requests to the store
that contain arrays in the query string.

For example:

    GET /segments?mapIds[]=id1&mapIds[]=id2&tags[]=t1&tags[]=t2


stephan83:
For compatibility with https://github.com/stratumn/sdk/pull/187

PR #23, opened 2017-10-09
agent-ui: create project

t-bast:
We want to re-write our agent-ui package in React/Redux (it was previously in ember).
We plan on using Material UI components (http://www.material-ui.com).
We'll use React Router and Immutable.


t-bast:
Do we need to setup right now:
- Travis CI
- ESLint
- Babel
- other stuff that should be done right from the beginning so it doesn't bite us later?
For each that we need to setup, can you let me know the recommendations?

jeremie-stratumn:
i don't think babel is required given react-scripts ..
we should definitely use travis and codecov. it is setup at the root of the project see: https://github.com/stratumn/indigo-js/blob/master/.travis.yml
So all we need is to setup a script for test:ci under packages/agent-ui/packages.json. 

I dont know how nice codecov plays with react-scripts. But for now we can just have test:ci point to test script.

such:
> Do we need to setup right now:

>- Travis CI
Yes: A minimal travis config will do I think:
```
language: node_js
node_js:
  - "stable"
cache:
  directories:
  - node_modules
script:
  - npm test
  - npm run build
```
>- ESLint
No (I will take care of it)
>- Babel`
No (Not needed for this project)

such:
You should prefix all of your commit with `agent-ui:` unless you want to squash and merge your pull request.
I'm not a big fan of Merge commits as well. Have you configured your git with:
```
[branch]
autosetuprebase = always
```

jeremie-stratumn:
if .travis.yml is defined in packages/agent-ui will it override the one at root level?

such:
@jeremie-stratumn My bad you're totally right. It should actually already work fine actually I think

jeremie-stratumn:
we just need a test:ci script I suppose..

t-bast:
Thanks for all the details, I'll address those.

Concerning the merge commit I'm setup with rebase but Github complained that this branch wasn't up-to-date with master and offered a nice "Update branch" button that I tried, and it looks like it created a merge commit from master to my branch.
I planned on squashing this initial one (which is why we didn't prefix with agent-ui:), but once this initial setup is committed we'll do small incremental PR with nicer commit messages :)

Do you know how I can get rid of the codeclimate error (which complains about the ... feature)? Is it an eslint config that we need to fix?

such:
It's related to your ESLint question actually I think. It seems that since we've not setup a .eslintrc file at the root of the project it runs default validation. We not the settle on the kind of style we want to use across project before this can work...

such:
@t-bast: I hate that update button! ;)

t-bast:
Makes sense. It's simply complaining that we haven't allowed the "..." ES feature yet, we can probably ignore it for this PR until we setup eslint for all projects?

such:
Yeah it's safe to ignore it. I've disabled the Github PR integration for now.

PR #24, opened 2017-10-10
react-mapexplorer: move package to indigo-js

such:
Used https://github.com/transitive-bullshit/react-modern-library-boilerplate to generate the project.


packages/react-mapexplorer/README.template.md
```diff
@@ -0,0 +1,31 @@
...
+    return (
+      <MyComponent />
+    )
+  }
+}
+```
+
+## License
+
+MIT © [Travis Fischer](https://github.com/transitive-bullshit)
```
t-bast: Is this file really needed? The example doesn't really use our mapexplorer component, does it?

packages/react-mapexplorer/example/README.md
```diff
@@ -0,0 +1,2164 @@
...
+Alternatively you may use `yarn`:
+
+```sh
+yarn add jest-enzyme
+```
+
+Import it in [`src/setupTests.js`](#initializing-test-environment) to make its matchers available in every test:
+
+```js
+import 'je
```
t-bast: Same, do we want to keep this?

packages/react-mapexplorer/example/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "react-mapexplorer-example",
+  "homepage": "https://stratumn.github.io/react-mapexplorer",
+  "version": "0.0.0",
+  "private": true,
+  "license": "MIT",
```
t-bast: Should we update this to Apache 2.0?

packages/react-mapexplorer/example/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "dependencies": {
+    "react": "^15.6.1",
+    "react-dom": "^15.6.1",
+    "react-scripts": "1.0.13"
```
t-bast: nit: react-scripts can be moved to dev-dependencies

packages/react-mapexplorer/README.template.md
such: not needed indeed!

packages/react-mapexplorer/example/README.md
such: not really

packages/react-mapexplorer/example/package.json
such: yeah it's not a big deal since we want to get rid of example as soon as agent-ui uses react-mapexplorer, but I'll change it in the meantime.

packages/react-mapexplorer/example/package.json
such: ok!

packages/react-mapexplorer/example/package.json
t-bast: Ok makes sense, no problem if you don't do it in this PR

PR #25, opened 2017-10-10
build: have a separate build for each package

t-bast:
It will make it easier to see what breaks when pushing changes (and a tiny bit faster since we'll have 5 parallel jobs instead of 1 sequential).
Renamed the agent-ui package for consistency with the other two packages.


such:
If #29 is merged first then in the root `package.json`, a new `scripts` will need to be added:
`"lint:ci": "eslint packages/$PACKAGE/src/** packages/$PACKAGE/test/**"`

`.travis.yml` will need to call `npm run lint:ci` rather than `npm run lint`

t-bast:
I'll let you merge #29 first and will update this PR afterwards

such:
For consistency we could have:
"test:ci": "lerna run --scope $PACKAGE test:ci" in root `package.json` and call `npm run test:ci` in `.travis.yml`.

t-bast:
That's exactly what I was doing right now :), thanks for confirming !

package.json
```diff
@@ -15,7 +15,9 @@
     "clean":
       "find packages -type d -name 'node_modules' -maxdepth 2 -exec rm -R {} \\;",
     "lint": "eslint packages/*/src/** packages/*/test/**",
+    "lint:ci": "eslint packages/$PACKAGE/src/** packages/$PACKAGE/test/**",
     "test": "lerna run test:ci",
```
t-bast: Should I change that to:
"test": "lerna run test" ?
I think it would be more logical

such:
;)
can you rebase master instead of merging? if it's not against anything you believe in!

t-bast:
Sure, no problem. I'll squash this one but will always rebase in the future :)

package.json
such: that would be more logical indeed but that doesn't work as is:
- `npm test` in `agent-ui ` and `mapexplorer-core` doesn't exit after the test so since it's not parallel one of them is never executed. We could have them executed in parallel but I'm afraid the output would be pretty messy...

PR #30, opened 2017-10-11
agent-ui: add node_modules to gitignore

jeremie-stratumn:
since root level has its own dev deps, we should ignore node_modules/ folder


such:
good catch! the commit message is wrong though ;)

PR #33, opened 2017-10-11
agent-js: support multiple evidences and fossilizers

### LINKED ISSUE #10: agent-js: allow multiple fossilizers
Fossilizers would be a priority because people might want to timestamp on different backends.

simonvadee:
Fix #10


packages/agent-js/src/process.js
```diff
@@ -14,44 +14,53 @@
...
       let callbackUrl = `${this.evidenceCallbackUrl || this.agentUrl}/${this
         .name}/evidence/${linkHash}`;
       callbackUrl += makeQueryString({ secret });
-      return this.fossilizerClient
-        .fossilize(linkHash, callbackUrl)
-        .then(() => segment);
+
+      this.pendingEvidences[linkHash] = this.fossilizerClients.length;
+      this.fossilizerClients.map(fossilizer =>
+        fossilizer.fossilize(linkHash, callbackUrl).then(() => segment)
```
t-bast: Does this execute the fossilize operations in parallel? If not is it simple to do?

packages/agent-js/src/process.js
```diff
@@ -79,15 +88,19 @@ export default class Process {
   * @returns {Promise} - a promise that resolve with the information
   */
   getInfo() {
-    const { processInfo } = this;
+    const processInfo = this.processInfo;
```
t-bast: nit: I believe the linter should keep the previous version of that line, are you sure it's correctly configured?

packages/agent-js/test/create.js
```diff
@@ -145,169 +156,188 @@ describe('Agent', () => {
...
         })
         .then(sgmts2 => {
           foundSegments.concat(sgmts2).length.should.be.exactly(2);
           callCount.should.be.exactly(2);
         });
     });
+
+    it("supports one or several fossilizers", () => {
+      const p1 = agent.addProcess("first", actions, memoryStore(), null);
+      should(p1.fossilizerClients).equal(null);
```
t-bast: very nit: p1.fossilizerClients.should.equal(null); is more consistent with how it's used in the rest of the test (if that syntax works)

packages/agent-js/test/httpServer.js
```diff
@@ -137,7 +137,7 @@ describe('HttpServer()', () => {
         segment.link.state.should.deepEqual({ a: 1, b: 2, c: 3 });
         segment.link.meta.mapId.should.be.a.String();
         segment.meta.linkHash.should.be.exactly(hashJson(segment.link));
-        segment.meta.evidence.should.deepEqual({ state: 'QUEUED' });
+        segment.meta.evidences.should.deepEqual([]);
```
t-bast: Maybe you can also validate the number of pending evidences in the pendingEvidences[linkHash] (if you have access to it here) ?

packages/agent-js/test/create.js
simonvadee: I agree it'd be more consistent, but if an object is null, it does not have a `should()` method

t-bast: Damn JS sucks :)

packages/agent-js/src/process.js
simonvadee: since the fossilizing isn't done on the agent side, it's not really parallelism. However the requests are sent in an asynchronous way and the `fossilizeSegment()` function returns before the fossilizers responded.

packages/agent-js/src/process.js
simonvadee: I changed it for more clarity

packages/agent-js/src/process.js
```diff
@@ -14,44 +14,53 @@
...
     this.actions = actions;
     this.storeClient = storeClient;
-    this.fossilizerClient = fossilizerClient;
-    this.salt = opts.salt || '';
+
+    if (fossilizerClients instanceof Object) {
+      this.fossilizerClients =
+        fossilizerClients instanceof Array
+          ? fossilizerClients
+          : [fossilizerClients];
```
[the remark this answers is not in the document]
simonvadee: you're right, javascript doesn't support function overloading. Forcing the input to be an array was a viable option but (after checking with @such) we decided to go for retro-compatibility and a more user-friendly interface (passing an array of one element doesn't make much sense). In the documentation though, the `fossilizerClients` argument is described as an array

packages/agent-js/test/httpServer.js
simonvadee: yes ! thank's

packages/agent-js/src/process.js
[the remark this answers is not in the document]
t-bast: Ok I can understand the need for the backwards compat', makes sense

packages/agent-js/src/create.js
```diff
@@ -14,9 +14,9 @@
   limitations under the License.
 */
 
-import httpServer from './httpServer';
-import Process from './process';
-import getDefinedFilters from './getDefinedFilters';
+import httpServer from "./httpServer";
+import Process from "./process";
+import getDefinedFilters from "./getDefinedFilters";
```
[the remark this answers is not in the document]
such: actually you should only need to set up `"prettier.eslintIntegration": true` so that it reads the eslint configuration file

packages/agent-js/src/process.js
```diff
@@ -23,15 +23,20 @@ import makeQueryString from './makeQueryString';
...
-
 export default class Process {
-  constructor(name, actions, storeClient, fossilizerClient, opts = {}) {
+  constructor(name, actions, storeClient, fossilizerClients, opts = {}) {
     this.name = name;
     this.actions = actions;
     this.storeClient = storeClient;
-    this.fossilizerClient = fossilizerClient;
+
+    if (fossilizerClients instanceof Object) {
```
such: is this check really needed?

such: maybe `if (fossilizerClients)` is enough?

packages/agent-js/src/process.js
```diff
@@ -40,18 +45,22 @@ export default class Process {
...
       pluginsInfo: getPluginsInfo(this.plugins)
     };
+    this.pendingEvidences = {};
   }
 
   fossilizeSegment(segment) {
-    if (this.fossilizerClient) {
-      const { linkHash } = segment.meta;
+    if (this.fossilizerClients) {
+      const linkHash = segment.meta.linkHash;
```
such: destructuring style should be enforced by eslint

simonvadee: it is now!

simonvadee:
The review for agent-js still stand (I did not commit any more updates), but I added the multi-evidences support for (react & angular)-mapexplorer

such:
I'm realizing that a commit is missing from the angular-mapexplorer project. I had started working on evidence customization which should help on this.

such:
@simonvadee see #56.
Conflicts ahead... Sorry for that.

packages/angular-mapexplorer/src/MapExplorer.controller.js
```diff
@@ -26,6 +26,18 @@ export default class MapExplorer {
...
+              evidence.date = new Date(
+                evidence.proof.timestamp * 1000
+              ).toUTCString();
+            } else if (evidence.backend === 'TMPop') {
+              evidence.merklePathString = JSON.stringify(
+                evidence.proof.original.merkleProof.InnerNodes
+              );
+            }
+            return evidence;
+          });
```
such: could this code be moved to the respective directive ?

PR #44, opened 2017-10-12
angular-mapexplorer: configure coverage with karma

such:


such:
@t-bast: nyc only works for code ran by Node. We use istanbul (and karma) to instrument (and test) code that is run in a browser.

t-bast:
Thanks for the explanation !

PR #45, opened 2017-10-13
agent-ui: add the process info page

t-bast:
Also in this PR:
* changed a bit the way we test components based on community best practices: I was previously kind of testing that redux worked well (using the store to dispatch props) whereas I can simplify this by testing only the components in isolation and trust that redux works
* moved the call to get agent info to the initialization (otherwise it wasn't executed when a user directly goes to a process url such as http://localhost:4000/myprocess)

Not in this PR but will come in a separate one:
* I'll add some tests for mapStateToProps
* I'll test the sub-components of the process info page (StoreSection, ActionsSection, FossilizersSection)

### DIFF
--- a/packages/agent-ui/src/components/agentInfoPage.js
+++ b/packages/agent-ui/src/components/agentInfoPage.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
-class AgentInfoPage extends Component {
+export class AgentInfoPage extends Component {
   render() {
     return (
       <div>
--- a/packages/agent-ui/src/components/processInfoPage.js
+++ b/packages/agent-ui/src/components/processInfoPage.js
@@ -1,0 +1,144 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+export class ProcessInfoPage extends Component {
+  render() {
+    const process = this.props.process;
+    if (!process || !process.name) {
+      return <div>Loading...</div>;
+    }
+
+    return (
+      <div>
+        <h1>{process.name}</h1>
+        <ActionsSection actions={process.processInfo.actions} />
+        <hr />
+        <StoreSection storeAdapter={process.storeInfo.adapter} />
+        <hr />
+        <FossilizersSection fossilizers={process.fossilizersInfo} />
+      </div>
+    );
+  }
+}
+
+function mapStateToProps(state, ownProps) {
+  let process = {};
+  if (state.agentInfo) {
+    if (state.agentInfo.processes[ownProps.params.process]) {
+      process = state.agentInfo.processes[ownProps.params.process];
+    }
+  }
+  return { process };
+}
+
+ProcessInfoPage.propTypes = {
+  process: PropTypes.object.isRequired
+};
+
+export const StoreSection = ({ storeAdapter }) => (
+  <div>
+    <h3>Store</h3>
+    <p>
+      <small>
+        A store is responsible for saving your data. There are different
+        adapters available depending on your needs.
+      </small>
+    </p>
+    <h4>Store adapter name</h4>
+    <samp>{storeAdapter.name}</samp>
+    <h4>Store adapter version</h4>
+    <samp>{storeAdapter.version}</samp>
+    <h4>Store adapter commit</h4>
+    <samp>{storeAdapter.commit}</samp>
+    <h4>Store adapter description</h4>
+    <samp>{storeAdapter.description}</samp>
+  </div>
+);
+
+StoreSection.propTypes = {
+  storeAdapter: PropTypes.object.isRequired
+};
+
+function formatActionSignature(action, args) {
+  let signature = action + '(';
+  for (let i = 0; args && i < args.length; ++i) {
+    signature += args[i];
+    if (i < args.length - 1) {
+      signature += ', ';
+    }
+  }
+  signature += ')';
+  return signature;
+}
+
+export const ActionsSection = ({ actions }) => (
+  <div>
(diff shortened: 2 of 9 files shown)


packages/agent-ui/src/components/agentInfoPage.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
-class AgentInfoPage extends Component {
+export class AgentInfoPage extends Component {
```
jeremie-stratumn: How does this export play with the other export at the bottom of the page? This might lead to including the wrong class down the line by mistake.. Could we export it with a different name?

packages/agent-ui/src/components/processInfoPage.js
```diff
@@ -0,0 +1,144 @@
+import React, { Component } from 'react';
```
jeremie-stratumn: I'd recommend to split this components into 4 components

packages/agent-ui/src/components/processInfoPage.js
```diff
@@ -0,0 +1,144 @@
...
+  signature += ')';
+  return signature;
+}
+
+export const ActionsSection = ({ actions }) => (
+  <div>
+    <h3>Actions</h3>
+    <ul>
+      {Object.keys(actions).map(action => (
+        <li key={action}>
```
jeremie-stratumn: Not necessary for now but you should look at material ui for list and other html tags.. They probably have nice built ins.

packages/agent-ui/src/setupTests.js
```diff
@@ -0,0 +1,4 @@
+import { configure } from 'enzyme';
```
jeremie-stratumn: can this live in test/ folder?

packages/agent-ui/src/store/dom.js
```diff
@@ -5,12 +5,16 @@ import { Provider } from 'react-redux';
 import { Router, browserHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
 
+import { getAgentInfo } from '../actions';
 import routes from '../routes';
 import { configureStore } from './';
 
 export default function(doc) {
   const store = configureStore();
+  store.dispatch(getAgentInfo());
```
jeremie-stratumn: ok for now, but I don't think this belongs here.. I think it should be the responsability of a component to dispatch actions. When we have multiple agents we'll have to dynamically getAgentInfo from a different url..

packages/agent-ui/src/components/agentInfoPage.js
t-bast: With that change, you can import the unconnected component via this file (import { AgentInfoPage } from './components/agentInfoPage'.
Otherwise if you import via the index.js (like we should do in non-test code) you can only import the default export (the connected component).
But I agree that it's clearer to rename this one to make it obvious that it's not connected, I'll look at what the community practices are for naming.

packages/agent-ui/src/components/processInfoPage.js
t-bast: I'm not a big fan of file explosion. The sub-components are made in React components because it's cleaner, but since they should only be used in this component I think it's better to keep them encapsulated in this file. If we want to reuse them in other files then it would make sense to move them to their own file.
Is that ok for you?

packages/agent-ui/src/components/processInfoPage.js
t-bast: Good point! I'll do another pass on that in another PR.

packages/agent-ui/src/setupTests.js
t-bast: Unfortunately no, the create-react scripts want to find this in src/setupTests.js.
But I've found one comment in a Github issue that was saying this might be configurable via package.json, I'll try that.

packages/agent-ui/src/store/dom.js
t-bast: Yeah I'm a bit unclear yet where this should ideally live.
But in App directly it didn't work for routes other than '/' :)
So we'll change that later when we need it, is that ok?

packages/agent-ui/src/components/agentInfoPage.js
jeremie-stratumn: you can rename on export I think. You could also have a indexTest.js at the root of components/ where you export and rename all the components for tests

packages/agent-ui/src/components/processInfoPage.js
jeremie-stratumn: your call. I'm pro code split because it is clearer IMO. I also think that react philosophy is small files and clear separation between components. It could also make sense to test those 4 pieces separately..

packages/agent-ui/src/store/dom.js
jeremie-stratumn: yes no pb

packages/agent-ui/src/components/agentInfoPage.js
t-bast: Actually even the redux documentation on testing connected components recommends exporting like this (with the same name): http://redux.js.org/docs/recipes/WritingTests.html
So I'll keep it this way, it will force us to import the connected component "properly" in prod code :)

packages/agent-ui/src/components/processInfoPage.js
t-bast: I will test them separately though in a next pull request, but I prefer encapsulation here (they are really not meant to be included anywhere else yet)

packages/agent-ui/src/setupTests.js
t-bast: Damn we can't, it's not supported by the react-scripts yet :(
Probably something they'll add/fix in later revisions

PR #48, opened 2017-10-13
ci: fix lint tests

such:
some packages have a different name than the directory they are in and they were missed by the previous lint:ci script.

We could have renamed the directories/package but I think this solution is not too ugly. Let me know what you think.


such:
@t-bast can you maybe fix it on this branch?

t-bast:
Definitely, ongoing

PR #51, opened 2017-10-13
agent-ui: add more unit tests

t-bast:
Add more unit tests for the process section component
Add data builders for test objects

### DIFF
--- a/packages/agent-ui/src/test/builders/index.js
+++ b/packages/agent-ui/src/test/builders/index.js
@@ -1,4 +1,2 @@
-/* eslint-disable */
-export { default as TestProcessBuilder } from "./testProcessBuilder";
-export { default as TestStateBuilder } from "./testStateBuilder";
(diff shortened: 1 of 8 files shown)


packages/agent-ui/src/test/builders/index.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable */
+export { default as TestProcessBuilder } from "./testProcessBuilder";
+export { default as TestStateBuilder } from "./testStateBuilder";
+/* eslint-enable */
```
such: I've just tested and you should not need the eslint-disable statement.
And it'll make you use single quotes ;)

t-bast: I'll try disabling it and pushing then, thanks !

t-bast: Done, works fine :)

PR #53, opened 2017-10-16
agent-js: drop legacy API support

such:
Does not make sense to keep them since with we have dropped compatibility anyway with the multi-process feature.


such:
@t-bast actually I think it's right in this case but this is not the right PR to fix those! :)

PR #54, opened 2017-10-16
agent-client-js: add promise getProcess function

### LINKED ISSUE #37: agent-client-js: add promise getProcess function
### Expected behavior
```javascript
getAgent(url)
  .then(agent => agent.getProcess(name))
  .then(process => process.doSomething())
  .catch(err => ...)
```
### Actual behavior
```javascript
getAgent(url)
  .then(agent => {
    const process = agent.processes[name];
    if (!process) throw new Error(...)
    return process.doSomething();
  })
  .catch(err => ...)
```

alexppxela:
Resolves #37

### DIFF
--- a/packages/agent-client-js/src/fromSegment.js
+++ b/packages/agent-client-js/src/fromSegment.js
@@ -16,14 +16,13 @@
 
 import getAgent from './getAgent';
 import getAdaptor from './getAdaptor';
-import getProcess from './getProcess';
 import segmentify from './segmentify';
 
 export default function fromSegment(obj) {
   return getAgent(
     obj.meta.agentUrl || obj.meta.applicationLocation
   ).then(agent => {
-    const process = getProcess(agent, obj.link.meta.process);
+    const process = agent.getProcess(obj.link.meta.process);
     const adaptor = getAdaptor(agent.url);
     const segment = segmentify(adaptor, process, obj);
     return { process, segment };
--- a/packages/agent-client-js/src/fromSegment.js
+++ b/packages/agent-client-js/src/fromSegment.js
@@ -20,13 +20,12 @@ import getProcess from './getProcess';
 import segmentify from './segmentify';
 
 export default function fromSegment(obj) {
-  return getAgent(obj.meta.agentUrl || obj.meta.applicationLocation)
-    .then(agent =>
-      Promise.all([agent, getProcess(agent, obj.link.meta.process)])
-    )
-    .then(([agent, process]) => {
-      const adaptor = getAdaptor(agent.url);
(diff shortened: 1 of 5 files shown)


such:
I think it is pretty confusing since there already exists a function `agent.getProcess()` (defined in `create.js`).

I'd find it a little weird to return a Promise from a synchronous function and we can actually re-write @jeremie-stratumn example like:

```
getAgent(url)
  .then(agent => Promise.resolve(agent.getProcess(name)))
  .then(process => process.doSomething())
  .catch(err => ...)
```

jeremie-stratumn:
I don't think agentClient offers a getProcess interface.. See https://github.com/stratumn/indigo-js/blob/37a1efbf74d8c4e62291657e00ab7c213e27a80b/packages/agent-client-js/src/getAgent.js#L27

So when using agentClient, you must `getAgent(url).then(agent => agent.processes.process) ...

such:
@jeremie-stratumn my bad I was confused indeed.
I think my point kinda still stands though and we should have:

```
const getProcess = (name) => {
  const process = agent.processes[name];
  if (!process) {
    throw new Error(`process '${name}' not found`);
  }
  return process;
}
```
bound somehow to the agent instance.

jeremie-stratumn:
Yeah I agree. It is weird to return a promise from a sync function.. And actually, once we have the getProcess function you mention above, we can easily do:
`getAgent(url).then(agent => agent.getProcess(name)).then(process => process.doSmthg())`
No need to artificially create a promise.. Thanks for your input.
@alexppxela sorry that I led you in the wrong direction here with the Promise route..

alexppxela:
No problem, all your comments are good to listen to be more efficient in JS.
I rework on this enh

packages/agent-client-js/src/fromSegment.js
```diff
@@ -16,21 +16,16 @@
...
 import segmentify from './segmentify';
 
 export default function fromSegment(obj) {
   return getAgent(
     obj.meta.agentUrl || obj.meta.applicationLocation
   ).then(agent => {
-    if (!agent.processes[obj.link.meta.process]) {
-      throw new Error(`process '${obj.link.meta.process}' not found`);
-    }
+    const process = getProcess(agent, obj.link.meta.process);
```
such: maybe we can also add
```
agent.getProcess = getProcess.bind(null, agent);
```
in `getAgent()` and replace this call by:
```
const process = agent.getProcess(obj.link.meta.process);
```

alexppxela: You are right, it is better.

PR #55, opened 2017-10-16
agent-js: track fossilizer and store clients created

t-bast:
This is a first step to support dynamically adding new processes to a running agent.
The agent needs to expose the fossilizers and stores that it already knows. This way the new process that will be added can either directly use one of those or supply new { url, name } objects to use.
I'm really not sure this code is the best way to achieve this, don't hesitate to suggest something completely different or tell me if you don't like it :)

### DIFF
--- a/packages/agent-js/src/fossilizerHttpClient.js
+++ b/packages/agent-js/src/fossilizerHttpClient.js
@@ -27,9 +27,9 @@ import handleResponse from './handleResponse';
  */
 export default function fossilizerHttpClient(url, opts = {}) {
   if (
-    fossilizerHttpClient.availableFossilizers.filter(
+    !fossilizerHttpClient.availableFossilizers.find(
       fossilizer => fossilizer.url === url
-    ).length === 0
+    )
   ) {
     fossilizerHttpClient.availableFossilizers.push({
       name: opts.name,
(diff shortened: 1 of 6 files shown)


packages/agent-js/src/fossilizerHttpClient.js
```diff
@@ -22,9 +22,21 @@ import handleResponse from './handleResponse';
...
  * @param {function} [opts.callbackUrl] - builds a URL that will be called with the evidence
+ * @param {string} [opts.name] - the name of the fossilizer
  * @returns {Client} a fossilizer HTTP client
  */
 export default function fossilizerHttpClient(url, opts = {}) {
+  if (
+    fossilizerHttpClient.availableFossilizers.filter(
+      fossilizer => fossilizer.url === url
+    ).length === 0
+  ) {
```
such: I think it's slightly better to write:
`if (fossilizerHttpClient.availableFossilizers.find(fossilizer => fossilizer.url === url))`

packages/agent-js/src/storeHttpClient.js
```diff
@@ -32,9 +32,21 @@ import filterAsync from './filterAsync';
...
+ * @param {object} [opt] - options
+ * @param {string} [opt.name] - the name of the store
  * @returns {Client} a store HTTP client
  */
-export default function storeHttpClient(url) {
+export default function storeHttpClient(url, opt = {}) {
+  if (
+    storeHttpClient.availableStores.filter(store => store.url === url)
+      .length === 0
+  ) {
```
such: same here

packages/agent-js/src/fossilizerHttpClient.js
t-bast: I'm a pure JS noob, thanks for pointing this out :)

PR #56, opened 2017-10-16
angular-mapexplorer: allow evidence customization

such:


packages/angular-mapexplorer/README.md
```diff
@@ -26,32 +26,51 @@ You might also need `tinycolor.js` and `angular-drop`.
...
+By default the directive expects the evidence to come from the Bitcoin Fossilizer.
+To override the evidence body you can transclude your own evidence component. The evidence part of the segment will be available on `$parent.evidence`
+
+```html
+<st-map-explorer chainscript="home.json" refresh="0">
+  <evidence-body>
+    <h1>State</h1>
+    <p>{{$parent.evidence.state}}</p>
+  </evidence-body>
+</st-map-explorer>
```
simonvadee: @such I wonder how this will fit with multiple evidences from different backends. Maybe the custom evidence component can be used when the backend is unknown ? In addition, a prop could be added in case we want to override some backend. What do you think ?

such: I don't think that's something you actually reuse directly in your PR. I think we'll actually have to drop this customization possibility and always display a "multi-evidence" directive.
This multi evidence directive will display a dedicated directive for each evidence.
NB: contrary to ember and React, it is not possible to dynamically include a directive in angular, we'll have to do a switch/case control structure.

PR #57, opened 2017-10-17
lint: install missing jslint dependency

t-bast:
VS Code is complaining on a clean clone of the repo that jslint is missing.
Adding it fixes editor issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-prettier": "^2.3.1",
     "eslint-plugin-react": "^7.4.0",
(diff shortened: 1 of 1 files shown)


such:
Isn't it because you have the jslint plugin activated in VSCode?
Since we don't use jslint for this project I'd rather not add it as a dependency. If you don't want your plugin to complain, you can either deactivate it or install jslint globally. Would that work for you? What do you think?

t-bast:
Good point, I'll check. I don't remember all the random stuff I installed initially without understanding if it would be useful or not :)

t-bast:
Yep it was indeed because I had the JSLint installed, conflicting with our ESLint. Thanks for clearing this mess :)

PR #60, opened 2017-10-17
agent-ui: migrate to react router v4 and material ui 1 beta

t-bast:


packages/agent-ui/src/actions/findSegments.js
```diff
@@ -0,0 +1,17 @@
+import { getAgent } from 'stratumn-agent-client';
+import { findSegmentsSuccess, findSegmentsFailure } from './';
+
+export default function(url, processName, options) {
+  return dispatch =>
+    getAgent(url)
+      .then(agent => {
+        const proc = agent.processes[processName];
```
jeremie-stratumn: we can now use alexandre's getProcess function: https://github.com/stratumn/indigo-js/commit/704914b032034bdf62d7fd517e8c1a249e62822e

packages/agent-ui/src/actions/getMapIds.js
```diff
@@ -0,0 +1,17 @@
+import { getAgent } from 'stratumn-agent-client';
+import { getMapIdsSuccess, getMapIdsFailure } from './';
+
+export default function(url, processName) {
+  return dispatch =>
+    getAgent(url)
+      .then(agent => {
+        const proc = agent.processes[processName];
```
jeremie-stratumn: same here, let's use getProcess https://github.com/stratumn/indigo-js/commit/704914b032034bdf62d7fd517e8c1a249e62822e

packages/agent-ui/src/components/processInfoPage.js
```diff
@@ -19,10 +19,11 @@ export const ProcessInfoPage = ({ process }) => (
 );
 
 export function mapStateToProps(state, ownProps) {
+  const processName = ownProps.match.params.process;
```
jeremie-stratumn: You are missing a `withRouter` at the bottom of the page here:
`export default withRouter(connect(mapStateToProps)(ProcessInfoPage));`

t-bast: Nope I don't need it here. You only need withRouter if you're not rendered inside a route, and the process info page is always rendered inside a route so it has the routing info without needing withRouter

packages/agent-ui/src/actions/findSegments.js
t-bast: Good idea, on it

PR #62, opened 2017-10-18
build: use a single travis build for all packages

t-bast:
There are a few issues with using separate builds per package in the mono-repo:
- it looks like sometime the mapexplore-core build simply didn't start at all (Travis throttling us?)
- most of the time the builds were run sequentially, not in parallel, so the overall build time was much longer
- most of the build time is spent in the bootstrapping part anyway, so it's a bit of a waste to do that in each separate build

Most of these issues are probably because we're on the public (free) Travis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
     "bootstrap": "lerna bootstrap",
     "clean": "lerna clean",
     "lint": "eslint packages/*/src packages/*/test",
-    "lint:ci": "lerna exec --scope $PACKAGE -- \\$LERNA_ROOT_PATH/node_modules/.bin/eslint \\$\\(pwd\\)/src \\$\\(pwd\\)/test",
+    "lint:ci": "eslint packages/*/src packages/*/test",
     "test": "lerna run test:ci",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,9 +14,9 @@
     "bootstrap": "lerna bootstrap",
     "clean": "lerna clean",
     "lint": "eslint packages/*/src packages/*/test",
-    "lint:ci": "lerna exec --scope $PACKAGE -- \\$LERNA_ROOT_PATH/node_modules/.bin/eslint \\$\\(pwd\\)/src \\$\\(pwd\\)/test",
+    "lint:ci": "eslint packages/*/src packages/*/test",
     "test": "lerna run test:ci",
-    "test:ci": "lerna run --scope $PACKAGE test:ci",
+    "test:ci": "lerna run test:ci",
```
such: `lint:ci` and `test:ci` are a bit useless now. If we decide to keep them, they should at least call `npm run lint` and `npm run test`...

t-bast: Yeah I was hesitating to keep them, maybe it's worth removing them entirely?

such: I think so too

PR #63, opened 2017-10-18
agent-ui: material-ui and align components 

t-bast:
I really think I'm doing it wrong so I'm creating this PR to gather some feedback from experts :).
In my opinion we shouldn't have to specify explicit width / margins and components should flow and align themselves correctly (but it's not the case yet).
@clinejm jérémie told me you have experience with React / Material UI, do you know how we're supposed to style our components correctly?

### DIFF
--- a/packages/agent-ui/src/components/topBar.js
+++ b/packages/agent-ui/src/components/topBar.js
@@ -7,6 +7,9 @@ import AppBar from 'material-ui/AppBar';
 import Toolbar from 'material-ui/Toolbar';
 import Typography from 'material-ui/Typography';
 
+import { withStyles } from 'material-ui/styles';
+import AppStyle from '../style/app';
+
 const Title = ({ urlPath }) => (
   <Typography type="title" noWrap>
     {urlPath}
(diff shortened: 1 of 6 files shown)


packages/agent-ui/src/components/topBar.js
```diff
@@ -19,34 +17,28 @@ Title.propTypes = {
...
-    width: `calc(100% - ${drawerWidth}px)`,
-    marginLeft: drawerWidth
-  }
-});
-
-const TopBar = ({ style, urlPath }) => (
-  <AppBar className={style}>
+const TopBar = ({ urlPath }) => (
+  <AppBar
+    style={{
```
clinejm: withStyle is added to the class, and there and styles.appBar has a similar declaration. should this be moved out to styles object?

t-bast: I removed withStyle from the class because for our simple usecases I thought inline css was smaller and simpler. If I understood correctly withStyle would be useful to access inherited css properties from the material UI theme, which I currently don't need. But I don't know what the right solution is :/

oliviertassinari:
> Looks good to me. I'm not sure what the best use of withStyles vs inline style objects.

@clinejm As a rule of thumbs, I only use inline-style for dynamics style properties. The CSS alternative provides more advantages. To name some of those pros:
- auto-prefixing
- better debug
- allow media queries
- allow keyframes

PR #64, opened 2017-10-19
agent-ui: refactor action dispatch logic, add maps info page

jeremie-stratumn:
I can see my linter went crazy so I'll need to fix that. dont approve yet

### DIFF
--- a/packages/agent-ui/src/actions/getAgentInfo.js
+++ b/packages/agent-ui/src/actions/getAgentInfo.js
@@ -1,9 +1,15 @@
 import { getAgent } from 'stratumn-agent-client';
 import { getAgentInfoSuccess, getAgentInfoFailure } from './';
 
-export default function(url) {
-  return dispatch =>
-    getAgent(url)
-      .then(agent => dispatch(getAgentInfoSuccess(agent)))
+export default function(name) {
+  return (dispatch, getState) => {
+    const state = getState();
+    if (!state.agents[name]) {
+      return dispatch(getAgentInfoFailure(`No agent named ${name}`));
+    }
+    const { url } = state.agents[name];
+    return getAgent(url)
+      .then(agent => dispatch(getAgentInfoSuccess(name, url, agent)))
       .catch(err => dispatch(getAgentInfoFailure(err)));
+  };
 }
--- a/packages/agent-ui/src/actions/getMapIds.js
+++ b/packages/agent-ui/src/actions/getMapIds.js
@@ -5,6 +5,7 @@ export default function(url, processName) {
   return dispatch =>
     getAgent(url)
       .then(agent => {
+        console.log(agent);
         const proc = agent.getProcess(processName);
         return proc.getMapIds();
       })
--- a/packages/agent-ui/src/components/agentInfoPage.js
+++ b/packages/agent-ui/src/components/agentInfoPage.js
@@ -1,49 +1,67 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
 
-export const AgentInfoPage = ({ agentUrl }) => (
-  <div>
-    <p>
-      An agent executes the logic of your processes. A process is defined by a
-      set of actions that may be used in the workflow.<br />
-      An instance of a process is called a map. It contains the different steps
-      of the process, called segments.
-    </p>
-    <hr />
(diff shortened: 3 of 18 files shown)


t-bast:
I'll add specific comments but I have two general comments:
- this pull request does too many separate things: can you split it? For example adding the maps info page should probably be in a different pull request
- can you add unit tests as well? We need to start unit testing stuff soon, otherwise if we realize later that we made some things impossible to test we'll waste a lot of time re-designing the architecture

packages/agent-ui/src/actions/getAgentInfo.js
```diff
@@ -1,9 +1,15 @@
...
-    getAgent(url)
-      .then(agent => dispatch(getAgentInfoSuccess(agent)))
+export default function(name) {
+  return (dispatch, getState) => {
+    const state = getState();
+    if (!state.agents[name]) {
+      return dispatch(getAgentInfoFailure(`No agent named ${name}`));
+    }
+    const { url } = state.agents[name];
+    return getAgent(url)
```
t-bast: If I understand correctly, you're fetching the agentInfo (with a network call). But if this agent was already stored in the state, why do you need to fetch it again? We should use the one that's stored locally. I believe changes to a running agent are not very frequent, so by default we should only load agents once and maybe offer a refresh button later to allow fetching updated agent info.

packages/agent-ui/src/actions/getMapIds.js
```diff
@@ -5,6 +5,7 @@ export default function(url, processName) {
   return dispatch =>
     getAgent(url)
       .then(agent => {
+        console.log(agent);
```
t-bast: nit: you can disagree with me on that :), but I think we shouldn't check-in code that does console.log. It's polluting the console and only useful when you're debugging a specific problem (and I tend to use the debugger instead of console logging) so it should rather be added only locally when testing out new stuff.

packages/agent-ui/src/components/agentInfoPage.js
```diff
@@ -1,49 +1,54 @@
...
+        <hr />
+      </div>
+    );
   }
-  return { agentUrl };
 }
 
-AgentInfoPage.propTypes = {
-  agentUrl: PropTypes.string.isRequired
+const mapStateToProps = ({ agents }, { match }) => {
```
t-bast: nit: rename "agents" to "agentsUrls"?

packages/agent-ui/src/components/app.js
```diff
@@ -1,29 +1,48 @@
...
+    if (this.props.match.params.agent !== match.params.agent)
+      dispatch(getAgentInfo(match.params.agent));
+  }
+
+  render() {
+    return (
+      <div>
+        <TopBar />
+        <LeftDrawer />
+        <InfoPage />
```
t-bast: This is a good idea: the app should only know that there are three components (drawer, topbar and info page) and info page should render the right thing depending on the route. I like it!

packages/agent-ui/src/components/app.js
```diff
@@ -1,29 +1,48 @@
...
 }
 
-App.propTypes = {
-  processes: PropTypes.arrayOf(PropTypes.string).isRequired
-};
+const AppContentConnected = withRouter(connect()(AppContent));
+
+const App = () => (
+  <Route
+    path="/:agent?/:process?/:detail?/:id?"
```
t-bast: I'm wondering how this can work.
In case I don't specify the agent and navigate to http://localhost:4000/myAwesomeProcess how can it know that this is a process and not an agent?
I think that this routing schema unfortunately can't work and will be a nightmare to maintain...

packages/agent-ui/src/components/app.js
```diff
@@ -1,29 +1,48 @@
...
+    component={AppContentConnected}
+  />
+);
 
-export default withRouter(connect(mapStateToProps)(App));
+export default withRouter(connect()(App));
```
t-bast: How does the component trigger the componentWillReceiveProps method if it's not connected to state changes? Only via route changes?

alexppxela:
If commits are distinct, it does not matter if the two functionalities are in the same PR. Commit history is the most important.

t-bast:
I don't agree with that because Github doesn't allow you to easily do commit reviews.
Since you do a review on a pull request, it's a lot easier and faster if pull requests are small and correctly scoped. 
Otherwise you honestly miss a lot of things, put too many comments with requires many new pushes and wastes a lot of time for everyone.

PR #65, opened 2017-10-20
agent-js: add new refs argument to createMap and createSegment

### LINKED ISSUE #12: agent-js: add new refs argument to createMap and createSegment
cf https://paper.dropbox.com/doc/Merge-Maps-yRLqBbu9jusqytgjvCTBD

alexppxela:
Resolves #12

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/65)

### DIFF
--- a/packages/agent-js/src/process.js
+++ b/packages/agent-js/src/process.js
@@ -141,14 +141,24 @@ export default class Process {
 
   /**
    * Creates the first segment of a map, calling the #init() function of the agent.
+   * @param {object[]} [refs] - the optional list of references of the new map
+   * @param {string} [refs.process] - ref process name from the same store
+   * @param {string} [refs.linkHash] - ref linkHash from the same store
+   * @param {object} [refs.segment] - ref linked segment from another store
+   * @param {string} [refs.meta] - ref type of relation
    * @param {...} args - the arguments to pass to the init function
    * @returns {Promise} - a promise that resolve with the segment
    */
-  createMap(...args) {
+  createMap(refs, ...args) {
     const initialLink = { meta: { mapId: uuid.v4() } };
     let link;
     let segment;
-    return processify(this.actions, initialLink)
+    return processify(
+      this.actions,
+      initialLink,
+      refs,
+      this.storeClient.getSegment
+    )
       .init(...args)
       .catch(err => {
         err.status = 400;
@@ -173,10 +183,15 @@ export default class Process {
    * Appends a segment to a map.
    * @param {string} prevLinkHash - the previous link hash
    * @param {string} action - the name of the transition function to call
+   * @param {object[]} [refs] - the optional list of references of the new segment
+   * @param {string} [refs.process] - ref process name from the same store
+   * @param {string} [refs.linkHash] - ref linkHash from the same store
+   * @param {object} [refs.segment] - ref linked segment from another store
+   * @param {string} [refs.meta] - ref type of relation
    * @param {...} args - the arguments to pass to the transition function
    * @returns {Promise} - a promise that resolve with the segment
    */
-  createSegment(prevLinkHash, action, ...args) {
+  createSegment(prevLinkHash, action, refs, ...args) {
     if (!this.actions[action]) {
       const err = new Error('not found');
       err.status = 404;
@@ -196,7 +211,12 @@ export default class Process {
       .then(() => {
         delete initialLink.meta.prevLinkHash;
         initialLink.meta.prevLinkHash = prevLinkHash;
-        const process = processify(this.actions, initialLink);
+        const process = processify(
+          this.actions,
+          initialLink,
+          refs,
+          this.storeClient.getSegment
+        );
         return process[action](...args).catch(err => {
           err.status = 400;
           throw err;
@@ -221,7 +241,7 @@ export default class Process {
    * Inserts evidence.
    * @param {string} linkHash - the link hash
    * @param {object} evidence - evidence to insert
-   * @param {strint} secret - a secret
+   * @param {string} secret - a secret
    * @returns {Promise} - a promise that resolve with the segment
    */
   insertEvidence(linkHash, evidence, secret) {
--- a/packages/agent-js/src/processify.js
+++ b/packages/agent-js/src/processify.js
@@ -16,7 +16,24 @@
 
 import protoChain from './protoChain';
 
-export default function processify(actions, initialLink) {
+function loadMeta(inputMeta, refs) {
+  const meta = JSON.parse(JSON.stringify(inputMeta || {}));
+  meta.refs = [];
+  if (refs == null) return meta;
+  if (!Array.isArray(refs)) {
+    throw new Error('Bad references type');
+  }
+
+  refs.forEach(ref => {
+    if ((!ref.process || !ref.linkHash) && ref.segment == null) {
+      throw new Error(`missing segment or (process and linkHash)`);
+    }
+    meta.refs.push(ref);
+  });
+  return meta;
+}
+
+export default function processify(actions, initialLink, refs, getSegment) {
   let link = JSON.parse(JSON.stringify(initialLink || {}));
 
   const methods = protoChain(
@@ -34,7 +51,7 @@ export default function processify(actions, initialLink) {
     new Promise((resolve, reject) => {
       const ctx = {
         state: JSON.parse(JSON.stringify(link.state || {})),
-        meta: JSON.parse(JSON.stringify(link.meta || {})),
+        meta: loadMeta(link.meta, refs),
         append(state, meta) {
           link = { state: state || this.state, meta: meta || this.meta };
           resolve(link);
@@ -44,6 +61,9 @@ export default function processify(actions, initialLink) {
           reject(
             err instanceof Error ? err : new Error(err || 'An error occurred')
           );
+        },
+        loadSegment(ref) {
+          return getSegment(ref.process, ref.linkHash);
         }
       };
 
--- a/packages/agent-js/test/httpServer.js
+++ b/packages/agent-js/test/httpServer.js
@@ -25,6 +25,7 @@ import plugins from '../src/plugins';
 import generateSecret from '../src/generateSecret';
 import actions from './utils/basicActions';
 import mockStoreHttpServer from './utils/mockStoreHttpServer';
+import refs from './utils/refs';
 
 function makePostRequest(server, url, args) {
   return supertest(server)
@@ -305,6 +306,7 @@ describe('HttpServer()', () => {
   describe('POST "/<process>/segments"', () => {
     it('renders the first segment', () => {
       const req = makePostRequest(server, `/${process.name}/segments`, [
+        [],
         1,
         2,
         3
@@ -323,9 +325,52 @@ describe('HttpServer()', () => {
       });
     });
 
+    it('renders the first segment with good refs', () => {
+      const req = makePostRequest(server, `/${process.name}/segments`, [
+        refs.getGoodRefs(process.name),
+        1,
+        2,
+        3
+      ]);
+      return testFn(req, (err, res) => {
+        if (err) {
+          throw err;
+        }
+        res.status.should.be.exactly(200);
+        const segment = res.body;
+        segment.link.state.should.deepEqual({ a: 1, b: 2, c: 3 });
+        segment.link.meta.mapId.should.be.a.String();
+        segment.link.meta.refs.should.be.an.Array();
+        segment.link.meta.refs.length.should.be.exactly(3);
+        segment.meta.linkHash.should.be.exactly(hashJson(segment.link));
+        segment.meta.evidences.should.deepEqual([]);
+        return should(process.pendingEvidences[segment.linkHash]).be.null;
+      });
+    });
+
+    it('renders the first segment with bad process ref', () => {
+      const req = makePostRequest(server, `/${process.name}/segments`, [
+        refs.getBadRefs(),
+        1,
+        2,
+        3
+      ]);
+      return testFn(req, (err, res) => {
+        if (err) {
+          throw err;
+        }
+        res.status.should.be.exactly(400);
+      });
+    });
+
     it('updates correclty when adding a process after initial launch', () => {
       const p2 = agent.addProcess('basic2', actions, memoryStore(), null);
-      const req = makePostRequest(server, `/${p2.name}/segments`, [1, 2, 3]);
+      const req = makePostRequest(server, `/${p2.name}/segments`, [
+        null,
+        1,
+        2,
+        3
+      ]);
       return testFn(req, (err, res) => {
         if (err) {
           throw err;
@@ -339,7 +384,7 @@ describe('HttpServer()', () => {
     });
 
     it('fails when trying to add segments to a non-existing process', () => {
-      const req = makePostRequest(server, '/lol/segments', [1, 2, 3]);
+      const req = makePostRequest(server, '/lol/segments', [[], 1, 2, 3]);
       return testFn(req, (err, res) => {
         if (err) {
           throw err;
@@ -351,11 +396,11 @@ describe('HttpServer()', () => {
 
   describe('POST "/<process>/segments/:linkHash/:action"', () => {
     it('renders the new segment', () =>
-      process.createMap(1, 2, 3).then(segment1 => {
+      process.createMap([], 1, 2, 3).then(segment1 => {
         const req = makePostRequest(
           server,
           `/${process.name}/segments/${segment1.meta.linkHash}/action`,
-          [4]
+          [[], 4]
         );
         return testFn(req, (err, res) => {
           if (err) {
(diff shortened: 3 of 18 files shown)


packages/agent-js/src/processify.js
```diff
@@ -37,6 +37,22 @@ export default function processify(actions, initialLink) {
         meta: JSON.parse(JSON.stringify(link.meta || {})),
         append(state, meta) {
           link = { state: state || this.state, meta: meta || this.meta };
+          link.meta.refs = [];
```
jeremie-stratumn: I can see here that link.meta.refs is not available to the action in the call to `method[key].apply(ctx, args)` further down, as it is added after the call to `append`. It could be useful some times to be able to get info from refs when creating a new segment that has multiple parents, just like when we currently create a new segment we have access to this.meta where we can find some previous info from the only parent.

packages/agent-js/src/processify.js
```diff
@@ -37,6 +37,22 @@ export default function processify(actions, initialLink) {
...
+            refs.forEach(ref => {
+              if ((!ref.process || !ref.linkHash) && ref.segment == null) {
+                reject(
+                  new Error(
+                    `missing segment or (process and linkHash) in ${ref}`
+                  )
+                );
+              }
+              link.meta.refs.push(ref);
+            });
```
jeremie-stratumn: should we throw if the refs provided is not an object Array?

packages/agent-js/src/processify.js
```diff
@@ -16,7 +16,23 @@
 
 import protoChain from './protoChain';
 
-export default function processify(actions, initialLink) {
+function readReferences(input) {
+  if (input == null) return [];
+  if (Object.prototype.toString.call(input) !== '[object Array]') {
```
such: I think we can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

packages/agent-js/src/processify.js
```diff
@@ -16,7 +16,23 @@
...
+    throw new Error('Bad references type');
+  }
+
+  const refs = [];
+  input.forEach(ref => {
+    if ((!ref.process || !ref.linkHash) && ref.segment == null) {
+      throw new Error(`missing segment or (process and linkHash) in ${ref}`);
+    }
+    refs.push(ref);
```
such: we could maybe load the segments that are referenced by linkHash and process with something like
```
storeClient.getSegment(ref.process, ref.linkHash)
```
so that they are available in the actions

packages/agent-js/src/processify.js
alexppxela: I tried but I should have wrote a mistake... It works well!

packages/agent-js/src/processify.js
alexppxela: I think it should be better to check segment only in sdk

such: It's not just to check that the segment exists, it's especially to provide it to the actions functions.

packages/agent-js/src/processify.js
```diff
@@ -35,24 +51,10 @@ export default function processify(actions, initialLink, refs) {
       const ctx = {
         state: JSON.parse(JSON.stringify(link.state || {})),
         meta: JSON.parse(JSON.stringify(link.meta || {})),
+        refs: readReferences(refs),
```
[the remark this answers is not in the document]
such: that makes sense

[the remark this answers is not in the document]
alexppxela: It is more logical but `append()` uses meta defined by user by default. If client forgets to insert refs in link.meta, references will be lost.
Sorry, I never trust the user behavior, so I used to secure the behavior and data in my code.
Maybe deleting refs should be the wanted client behavior.

packages/agent-js/src/processify.js
alexppxela: @such Can you check if 28a19fe matches with your thoughts?

packages/agent-js/src/processify.js
```diff
@@ -50,18 +56,22 @@ export default function processify(actions, initialLink, refs) {
...
           resolve(link);
         },
         reject(error) {
           const err = error || this.error;
           reject(
             err instanceof Error ? err : new Error(err || 'An error occurred')
           );
+        },
+        loadSegment(ref) {
+          return loadedRefSegments({
```
jeremie-stratumn: I am confused.. It looks like you expect that `loadedRefSegments` is a function but you pass `this.loadRefSegments(refs)` to it with is not a function..

such: A test is probably missing as well :)

alexppxela: I'm writing it showing, it misses `.get`. I will test binding a function to load segment only on function call.

packages/agent-js/src/process.js
```diff
@@ -141,14 +141,24 @@ export default class Process {
...
+  createMap(refs, ...args) {
     const initialLink = { meta: { mapId: uuid.v4() } };
     let link;
     let segment;
-    return processify(this.actions, initialLink)
+    return processify(
+      this.actions,
+      initialLink,
+      refs,
+      this.storeClient.getSegment.bind()
```
such: I don't think it's useful to call `bind` here

packages/agent-js/test/httpServer.js
```diff
@@ -235,9 +237,46 @@ describe('HttpServer()', () => {
       });
     });
 
+    it('renders the first segment with good refs', () => {
+      const req = supertest(server)
```
jeremie-stratumn: can you use `makePostRequest` function?

packages/agent-js/test/utils/refs.js
```diff
@@ -0,0 +1,54 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+export default {
+  getGoodRefs(process, linkHash) {
+    const l =
+      linkHash ||
+      'fa871e81c469fa947eacd40f89dc5627a0cb3a96551a651c034787c752d4448f';
+    return JSON.parse(
```
jeremie-stratumn: Please use object literal here, no need to `JSON.parse`:
```javascript
return [
{
  process,
  linkhash: l,
...
},
...
]
```

packages/agent-js/test/httpServer.js
```diff
@@ -235,9 +237,46 @@ describe('HttpServer()', () => {
...
+        segment.meta.linkHash.should.be.exactly(hashJson(segment.link));
+        segment.meta.evidences.should.deepEqual([]);
+        return should(process.pendingEvidences[segment.linkHash]).be.null;
+      });
+    });
+
+    it('renders the first segment with bad process ref', () => {
+      const req = supertest(server)
+        .post(`/${process.name}/segments`)
+        .send([JSON.parse(`[{ "process": "foo"}]`), 1, 2, 3]);
```
jeremie-stratumn: use object literal. See later comment

packages/agent-js/test/process.js
```diff
@@ -160,8 +181,62 @@ describe('Process', () => {
...
+            );
+            segment2.meta.linkHash.should.be.exactly(hashJson(segment2.link));
+            segment2.link.meta.refs.should.be.an.Array();
+            segment2.link.meta.refs.length.should.be.exactly(3);
+          });
+      });
+    });
+
+    it('checks loadSegment function with good references', () => {
+      process.actions.test = function test() {
```
jeremie-stratumn: I'm not a big fan of setting this in the test body like this.. what if process.actions already have a `test` action?

packages/agent-js/test/process.js
```diff
@@ -160,8 +181,62 @@ describe('Process', () => {
...
+        }
+      };
+
+      return process.createMap([], 1, 2, 3).then(segment1 =>
+        process
+          .createSegment(
+            segment1.meta.linkHash,
+            'test',
+            refs.getGoodRefs(process.name, segment1.meta.linkHash),
+            4
```
jeremie-stratumn: what is this 4 for?

packages/agent-js/test/process.js
```diff
@@ -160,8 +181,62 @@ describe('Process', () => {
...
+          )
+          .then(segment2 => {
+            segment2.link.state.nbSeg.should.be.exactly(1);
```
jeremie-stratumn: So here you are testing the number of segments refs that you were able to load? Would be good to count the errors as well and test that. Could you add comments here to explain what you are doing?

packages/agent-js/src/process.js
```diff
@@ -196,7 +211,12 @@ export default class Process {
       .then(() => {
         delete initialLink.meta.prevLinkHash;
         initialLink.meta.prevLinkHash = prevLinkHash;
-        const process = processify(this.actions, initialLink);
+        const process = processify(
+          this.actions,
+          initialLink,
+          refs,
+          this.storeClient.getSegment.bind()
```
jeremie-stratumn: not sure why you need to bind ... ? Can't you just pass the `getSegment` function directly?

packages/agent-js/src/process.js
alexppxela: You are true. I've changed it, thanks

packages/agent-js/src/process.js
alexppxela: done

packages/agent-js/test/httpServer.js
alexppxela: sure

packages/agent-js/test/httpServer.js
alexppxela: I forgot your previous comment, it's now ok

packages/agent-js/test/process.js
alexppxela: I have left a parameter, I forgot that it is possible to ommit an argument in JS...
I have deleted it

packages/agent-js/test/process.js
alexppxela: I prefer like this because this functionality is only tested in that use case.
Pros are code to test is centralized in the function, cons is that test file is bigger.

packages/agent-js/test/utils/refs.js
```diff
@@ -19,36 +19,41 @@ export default {
...
-              "prevLinkHash": "2f0d1bde3f163446d6f23e31bd0b43063ec07ea4128cf5a258208e9f80832735",
-              "process": "foobar",
-              "stateHash": "124c9b6c55def7bc177c81d98ba12898b018a534c3101961a4b683678fa2d679"
-            }
-          },
-          "meta": {
-            "linkHash": "e312a9d43d1401eeb62af52810ee8b6b746b67835eb040755ccb1598a5c88411",
-            "segmentUrl": "http://localhost:3000/segments/e312a9d43d1401eeb62af52810ee8b6b746b67835eb040755ccb1598a5c88411"
+    const goodLink = {
+      process: process,
```
jeremie-stratumn: I am surprised that your linter did not pick that up. Not a big deal but FYI, when key and value are the "same" you can omit the value (ES6). So here no need to repeat `process`:
```
const goodLink = {
  process,
  linkHash: l,
  ...
}

alexppxela: Thanks for this advice. I resquash to prepare @simonvadee PR based on this branch.
Finally, I moved testLoadSegments in basicAgent.js

t-bast:
Reviewed 10 of 10 files at r1, 9 of 9 files at r2.

---

*[packages/agent-client-js/src/httpAdaptor.js, line 33 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/65#-KyMK0wshhGl46cP0S8a:-KyMK0wtZiczO_5VlX7n:b-g5ocfl) ([raw file](https://github.com/stratumn/indigo-js/blob/d48c96466a74482c132adfe3b212296274e7f227/packages/agent-client-js/src/httpAdaptor.js#L33)):*
> ```JavaScript
>   }
> 
>   createMap(processName, refs, ...args) {
> ```

It's really a pain that javascript doesn't have function overloading, it looks like each time we'll need to add a required parameter to those functions we will need to do a ton of updates (and will probably miss some since we can't statically check). 
In that case it's particularly annoying because refs isn't really required and most of the time will be [].
Do you think it would be better to replace the new "refs" argument by an "opts" object that would contain refs (and default to []) and will be helpful if later we need to add more arguments (this way we would add them inside the opts object itself and won't have to change the function signature everywhere) ? It's just a random thought seeing how painful it is to "simply" add this argument in the full codebase

---

*[packages/agent-client-js/src/segmentify.js, line 21 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/65#-KyMGzEqBbaphl3lP3EW:-KyMGzErCE7Xe8dEbSCm:bj1mvsv) ([raw file](https://github.com/stratumn/indigo-js/blob/d48c96466a74482c132adfe3b212296274e7f227/packages/agent-client-js/src/segmentify.js#L21)):*
> ```JavaScript
> export default function segmentify(adaptor, process, obj) {
>   Object.keys(process.processInfo.actions)
>     .filter(action => ['init'].indexOf(action) < 0)
> ```

hum, I'm probably missing something but why not simply action !== 'init' ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/65)*

such:
What's the status on this? Can the comments from @t-bast be addressed?

alexppxela:
*[packages/agent-client-js/src/httpAdaptor.js, line 33 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/65#-KyMK0wshhGl46cP0S8a:-Kz3IyPCVvJ8n7qKar3v:b7aalbe) ([raw file](https://github.com/stratumn/indigo-js/blob/d48c96466a74482c132adfe3b212296274e7f227/packages/agent-client-js/src/httpAdaptor.js#L33)):*

It's really a pain that javascript doesn't have function overloading, it looks like each time we'll need to add a required parameter to those functions we will need to do a ton of updates (and will probably miss some since we can't statically check). 
In that case it's particularly annoying because refs isn't really required and most of the time will be [].
Do you think it would be better to replace the new "refs" argument by an "opts" object that would contain refs (and default to []) and will be helpful if later we need to add more arguments (this way we would add them inside the opts object itself and won't have to change the function signature everywhere) ? It's just a random thought seeing how painful it is to "simply" add this argument in the full codebase
</blockquote></details>

I don't think that it is necessary to wrap `refs` in an `opts` object for the moment. You're true when you say that usual  value will be `[]` but it is easier to read than passing an object to parse to agent-js.
I think an `opts` structure can wrap `refs` later if we need another typed field

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/65)*

t-bast:

---

*[packages/agent-client-js/src/httpAdaptor.js, line 33 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/65#-KyMK0wshhGl46cP0S8a:-Kz3ptF4eFeK3j4dpKYY:bb4ffwz) ([raw file](https://github.com/stratumn/indigo-js/blob/d48c96466a74482c132adfe3b212296274e7f227/packages/agent-client-js/src/httpAdaptor.js#L33)):*
<details><summary><i>Previously, alexppxela (Alexandre Thibault) wrote…</i></summary><blockquote>

I don't think that it is necessary to wrap `refs` in an `opts` object for the moment. You're true when you say that usual  value will be `[]` but it is easier to read than passing an object to parse to agent-js.
I think an `opts` structure can wrap `refs` later if we need another typed field
</blockquote></details>

Ok sounds good, let's keep it in mind next time we need to add something in those args

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/65)*

alexppxela:
Review status: 12 of 19 files reviewed at latest revision, 1 unresolved discussion.

---

*[packages/agent-client-js/src/segmentify.js, line 21 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/65#-KyMGzEqBbaphl3lP3EW:-KzULWzyVPLFcRSq3Tm1:b-896fix) ([raw file](https://github.com/stratumn/indigo-js/blob/d48c96466a74482c132adfe3b212296274e7f227/packages/agent-client-js/src/segmentify.js#L21)):*

hum, I'm probably missing something but why not simply action !== 'init' ?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/65)*

such:
Reviewed 6 of 10 files at r1, 6 of 9 files at r2, 6 of 6 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/65)*

such:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/65#-:-KzUM82CTZoBNwW08UWr:bnfp4nl)*

PR #67, opened 2017-10-23
agent-ui: create skeleton app

t-bast:
This is unfortunately a very big pull request :( but a lot of it is boilerplate / almost empty react components.
Jérémie and I decided to re-start from scratch with what we learnt in our first run with React / Redux / React Router / a ton of other JS frameworks needed to create a clean and viable codebase nowadays.
The main mistake we made with what's currently in master is that we rushed a bit too much to get to feature parity with the ember agent-ui instead of thinking about the state and routing architecture of the app. We added styling and material-ui way too soon, and we realized too late that we had made a few choices that were slowing us down.
So we chose re-work on the foundations of the codebase: clean-up the dependencies and make sure we understood them, clean up the architecture to make the app easy to unit-test and evolve and focus on getting data to flow properly between pages before doing any styling / complex pages / components.

So I re-read the documentations of React, Redux, React-Router v4 and Enzyme, played with the Redux real-world code samples and re-build the project architecture by copying the best practices from the Redux codebase. 
With this pull request we have the routing architecture properly setup, state object is (almost) fully defined to handle our data, and a few unit tests have been implemented to validate that everything is properly decoupled and unit-testable.
The next steps are:
* merge this in master
* add more unit tests to cover the features already implemented
* finish cleaning up the "old" code (mainly what's inside the components folder)
* implement proper data flow to all future pages without any styling
* only then start adding new features and maybe style


packages/agent-ui/src/containers/agentInfoPage.js
```diff
@@ -0,0 +1,34 @@
...
+
+AgentInfoPage.propTypes = {
+  name: PropTypes.string.isRequired,
+  url: PropTypes.string.isRequired
+};
+
+function mapStateToProps(state, ownProps) {
+  console.log('AgentInfoPage state', state);
+  console.log('AgentInfoPage ownProps', ownProps);
+  const name = ownProps.match.params.agent;
```
jeremie-stratumn: I thought react-router-redux would provide all the location / match info via state, no?

packages/agent-ui/src/containers/agentsPage.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+AgentsPage.propTypes = {
+  fetchAgent: PropTypes.func.isRequired
+};
+
+function mapStateToProps(state, ownProps) {
+  console.log('AgentsPage state', state);
+  console.log('AgentsPage ownProps', ownProps);
+  return {};
```
jeremie-stratumn: We should probably list the agents already added on this page.. So we could extract them from state and return that.

packages/agent-ui/src/reducers/agents.js
```diff
@@ -0,0 +1,18 @@
...
+export default function(state = {}, action) {
+  const agent = {};
+  switch (action.type) {
+    case actionTypes.AGENT_INFO_REQUEST:
+      agent[action.name] = {
+        status: 'LOADING',
+        url: action.url,
+        processes: {}
+      };
+      return Object.assign({}, state, agent);
```
jeremie-stratumn: I like the spread operator syntax recommended by redux: http://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html
` return { ...state, agent)`

packages/agent-ui/src/reducers/agents.js
```diff
@@ -0,0 +1,18 @@
...
+    case actionTypes.AGENT_INFO_FAILURE:
+    case actionTypes.AGENT_INFO_SUCCESS:
+    default:
```
jeremie-stratumn: You want to implement the other cases later?

packages/agent-ui/src/containers/agentsPage.js
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <div>
+      <form
+        onSubmit={e => {
+          e.preventDefault();
+          if (!agentName.value.trim() || !agentUrl.value.trim()) {
+            return;
+          }
+          fetchAgent(agentName.value, agentUrl.value);
```
jeremie-stratumn: Not sure we want to fetch agent directly here.. I think we should update the dictionnary in the state that store the mapping name -> url and that's it. Only when user clicks on the link to go to the agent page we should fetchAgent..

packages/agent-ui/src/store/configure.js
```diff
@@ -1,12 +1,7 @@
...
 
-export default function() {
-  return createStore(
-    combineReducers({
-      agentInfo: getAgentInfo
-    }),
-    applyMiddleware(thunk)
-  );
-}
+const configureStore = () => createStore(rootReducer, applyMiddleware(thunk));
```
jeremie-stratumn: add redux-logger middleware too pls

packages/agent-ui/src/actions/getAgent.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+  afterEach(() => {
+    stratumnClientStub.restore();
+  });
+
+  it('calls stratumn client to get agent', () => {
+    stratumnClientStub.rejects('Unreachable');
+
+    return getAgent('', '')(dispatchSpy).then(() => {
+      expect(stratumnClientStub.callCount).to.equal(1);
```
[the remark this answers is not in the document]
t-bast: For the dispatchSpy the count is 2, but for the stratumnClientStub (which stubs the calls to stratumn-client's getAgent) the callCount should always be 1 (we only call getAgent once)

packages/agent-ui/src/actions/getMapIds.js
```diff
@@ -1,13 +1,33 @@
...
+
+const getMapIdsSuccess = mapIds => ({
+  type: actionTypes.MAP_IDS_SUCCESS,
+  mapIds
+});
+
+export default function(agent, process) {
+  return dispatch => {
+    dispatch(getMapIdsRequest(agent, process));
+    // TODO: get agent url
```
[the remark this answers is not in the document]
t-bast: Neat, I didn't want to handle that in this PR but it's good to know that it's going to be simple :)

packages/agent-ui/src/containers/agentInfoPage.js
t-bast: Good point, I need to update this to always look for it in state.

packages/agent-ui/src/containers/agentsPage.js
t-bast: If we do that that means we can't display the list of processes in the left nav until the user clicks on that agent...
From a user's perspective I find it more intuitive that when I click "Add Agent" after having entered the url it will effectively get that agent right now (but that's only my opinion).
I think both options are valid design choices, no need to commit to a specific choice right now, we can update this in another PR

packages/agent-ui/src/containers/agentsPage.js
t-bast: Yes definitely. But that's for another PR.

packages/agent-ui/src/reducers/agents.js
t-bast: Indeed, it's a much better syntax, will update that.

packages/agent-ui/src/reducers/agents.js
t-bast: Exactly, this PR is already huge whereas implementing this can be done in a small and easy-to-review dedicated PR.

packages/agent-ui/src/store/configure.js
t-bast: Will be done in another PR.

packages/agent-ui/src/actions/getAgent.test.js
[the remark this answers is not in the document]
jeremie-stratumn: right, my bad

packages/agent-ui/src/actions/getMapIds.js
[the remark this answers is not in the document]
jeremie-stratumn: yeah we can do that later

packages/agent-ui/src/containers/agentsPage.js
jeremie-stratumn: ok for another PR. But on the UX question, adding an agent can dispatch a sync action like `addAgent` and both this page and leftDrawer will know to update their props.. So both should render correctly

packages/agent-ui/src/containers/agentsPage.js
jeremie-stratumn: sure

packages/agent-ui/src/reducers/agents.js
jeremie-stratumn: sounds good

packages/agent-ui/src/store/configure.js
jeremie-stratumn: sure

packages/agent-ui/src/containers/agentInfoPage.js
t-bast: In fact the only thing that react-router-redux can do for us is to put the "location" part in the store. It cannot put the match part in the store (see this Github issue: https://github.com/ReactTraining/react-router/issues/5096). So we'll have to accept that react-router will provide this information in our ownProps and can't put it in the store's state...

PR #68, opened 2017-10-23
agent-js: API to add a new process

t-bast:
Adding a new API on the agent to allow dynamic upload of a new process (by script injection).
We most likely only want to allow this for test networks as an opt-in (it's way too dangerous otherwise) so I think I'll disable this API by default and allow enabling it (instead of having it enabled by default like what we have in this PR). What do you think?

Note: this is still lacking configuration of the store and fossilizer (hence the memorystore used and null fossilizer). This will come in another PR to keep this one small (since it raises a few issues, I'd like to focus on those before implementing store/fossilizer configuration which can also raise other issues).

### DIFF
--- a/packages/agent-js/src/httpServer.js
+++ b/packages/agent-js/src/httpServer.js
@@ -92,20 +92,13 @@ export default function httpServer(agent, opts = {}) {
           return res.status(400).json({ error: 'invalid script' });
         }
 
-        if (exported.name !== req.params.process) {
-          return res.status(400).json({
-            error:
-              "Process name from url doesn't match process name from the script"
-          });
-        }
-
         if (!exported.init || typeof exported.init !== 'function') {
           return res.status(400).json({
             error: 'missing init function'
           });
         }
 
-        agent.addProcess(exported.name, exported, memorystore(), null);
+        agent.addProcess(req.params.process, exported, memorystore(), null);
 
         return res.json(agent.getAllProcesses());
       } catch (err) {
--- a/packages/agent-js/src/httpServer.js
+++ b/packages/agent-js/src/httpServer.js
@@ -27,6 +27,7 @@ import memorystore from './memoryStore';
  * @param {object} [opts] - options
  * @param {object} [opts.cors] - CORS options
  * @param {object} [opts.salt] - salt used for callback URLs
+ * @param {object} [opts.enableProcessUpload] - allow clients to upload new processes to a running agent
  * @returns {express.Server} - an express server
  */
 export default function httpServer(agent, opts = {}) {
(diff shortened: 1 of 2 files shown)


such:
The opt-in configuration is a great idea I think.

t-bast:
Cool, I'll add a commit to this PR to set this up

packages/agent-js/src/httpServer.js
```diff
@@ -62,6 +63,52 @@ export default function httpServer(agent, opts = {}) {
...
+      if (!exported.init || typeof exported.init !== 'function') {
+        return res.status(400).json({
+          error: 'missing init function'
+        });
+      }
+
+      agent.addProcess(exported.name, exported, memorystore(), null);
+
+      return res.json(agent.getAllProcesses());
+    } catch (err) {
```
such: do you think we need this big wrapping try catch control? Shouldn't this be handled by our error express middleware?

packages/agent-js/src/httpServer.js
```diff
@@ -62,6 +63,52 @@ export default function httpServer(agent, opts = {}) {
...
+          Buffer.from(req.body.script, 'base64').toString(),
+          '' // this parameter is actually required (undefined isn't accepted)
+        );
+        /* eslint-enable no-underscore-dangle */
+        exported = processModule.exports;
+      } catch (err) {
+        return res.status(400).json({ error: 'invalid script' });
+      }
+
+      if (exported.name !== req.params.process) {
```
such: do we really need to enforce the presence of the name attribute in the action object if it is mandatory in the URL anyway? We can always add in ourselves if it only for "storage".

t-bast: Good point, initially I was using that along with nodemon to create a new file for the process that I was putting in /lib/actions/actions-{process}.js but since we shouldn't run nodemon in prod, I changed that to directly load the script in memory. I kept the name inside the script but it's not necessary anymore, will remove it.

packages/agent-js/src/httpServer.js
t-bast: I wanted to be able to provide a good error message that explained that the script was invalid, if I do that at the base error handler layer I'll lose some context information and won't be able to provide that, will I?

t-bast: In fact I'll clean up the whole file's usage of middlewares, but I think it's worth doing it in a separate PR, is that ok?

such: sure

packages/agent-js/src/httpServer.js
```diff
@@ -62,6 +64,49 @@ export default function httpServer(agent, opts = {}) {
...
 
+  /**
+   * This API allows dynamic upload of new processes to a running agent.
+   * This is obviously very dangerous as it accepts any javascript code,
+   * which is why it's disabled by default.
+   * If you choose to enable it, a strict access control mechanism should be 
+   * setup to restrict access to this API.
+   */
+  if (opts.enableProcessUpload) {
+    app.post('/:process/add', (req, res, next) => {
```
such: do we want to rename the api "upload" to be consistent with the option name?

t-bast: sounds, good, I'm on it

PR #69, opened 2017-10-23
agent-js: fix plugins filter tests

such:


packages/agent-js/test/process.js
```diff
@@ -280,40 +344,43 @@ describe('Process', () => {
...
       ];
 
-      process
+      return process
         .createMap(2, 2, 3)
-        .then(() => process.findSegments('test'))
+        .then(() => process.findSegments())
         .then(body => {
-          body.should.have.length(2);
+          body.should.have.length(1);
```
alexppxela: The test is more revealant now!

PR #70, opened 2017-10-23
agent-js: remove useless forwarding to error middleware

t-bast:
According to Express documentation, we only ever explicitly call the next handler in the middleware chain if our handler code has finished executing but the request isn't handled yet.
In our cases our handlers either correctly write the http response to the output buffer or an exception is thrown. So since we're using an error middleware to catch exceptions, there's no need to explicitly forward the request to it (we're duplicating a try/catch that Express already internally does).


t-bast:
Looks like this is causing issues in the travis build in agent-client-js tests, I'll investigate why (it doesn't seem to repro locally though)

t-bast:
In fact in handlers that directly return promises, it looks like we need our manual forwarding (or maybe we need to declare the app.use differently if we want Express to track our unresolved promise and use the middleware chain, I'll dig into their documentation to see if I can find something more elegant)

t-bast:
async-middleware to the rescue, exactly meant for that purpose ;)

PR #72, opened 2017-10-24
agent-js: process upload API accepts store and fossilizer configuration

t-bast:
Adding store and fossilizer configuration options to the process upload API.
Refactored the process upload API as well (extracted several smaller functions).
I have one open question about the evidence callback url: I'm currently forcing the caller to specify a callback url for each fossilizer. This is maybe a bit drastic, if none is specified I could use agent.agentUrl or something like it as a default value, what do you think?
Overall I found that the evidence callback url could be configured via a few different variables so I'm a bit unclear about this, maybe it needs a bit of refactoring?

### DIFF
--- a/packages/agent-js/src/fossilizerHttpClient.js
+++ b/packages/agent-js/src/fossilizerHttpClient.js
@@ -26,11 +26,8 @@ import handleResponse from './handleResponse';
  */
 export default function fossilizerHttpClient(url, opts = {}) {
   // If we already have an existing fossilizer for that url, re-use it
-  const matchingFossilizerClient = fossilizerHttpClient.availableFossilizers.find(
-    foss => foss.url === url
-  );
-  if (matchingFossilizerClient) {
-    return matchingFossilizerClient.client;
+  if (fossilizerHttpClient.availableFossilizers[url]) {
+    return fossilizerHttpClient.availableFossilizers[url];
   }
 
   const fossilizerClient = {
@@ -78,22 +75,21 @@ export default function fossilizerHttpClient(url, opts = {}) {
     }
   };
 
-  fossilizerHttpClient.availableFossilizers.push({
-    url: url,
-    client: fossilizerClient
-  });
+  fossilizerHttpClient.availableFossilizers[url] = fossilizerClient;
 
   return fossilizerClient;
 }
 
-fossilizerHttpClient.availableFossilizers = [];
+fossilizerHttpClient.availableFossilizers = {};
 
 /**
  * Returns basic information about the fossilizer HTTP clients that have been created.
- * @returns {Array} an array of fossilizer HTTP clients basic information
+ * @returns {Array} an array of fossilizer HTTP clients basic information (currently only url).
  */
 export function getAvailableFossilizers() {
-  return fossilizerHttpClient.availableFossilizers.map(f => ({ url: f.url }));
+  return Object.keys(fossilizerHttpClient.availableFossilizers).map(url => ({
+    url
+  }));
 }
 
 /**
@@ -101,5 +97,5 @@ export function getAvailableFossilizers() {
  * Should only be used for tests setup.
  */
 export function clearAvailableFossilizers() {
-  fossilizerHttpClient.availableFossilizers = [];
+  fossilizerHttpClient.availableFossilizers = {};
 }
--- a/packages/agent-js/src/fossilizerHttpClient.js
+++ b/packages/agent-js/src/fossilizerHttpClient.js
@@ -26,18 +26,19 @@ import handleResponse from './handleResponse';
  * @returns {Client} a fossilizer HTTP client
  */
 export default function fossilizerHttpClient(url, opts = {}) {
-  if (
-    !fossilizerHttpClient.availableFossilizers.find(
-      fossilizer => fossilizer.url === url
-    )
-  ) {
-    fossilizerHttpClient.availableFossilizers.push({
-      name: opts.name,
-      url: url
-    });
+  // If we already have an existing fossilizer for that url, re-use it
+  const matchingFossilizerClient = fossilizerHttpClient.availableFossilizers.find(
+    foss => foss.url === url
+  );
+  if (matchingFossilizerClient) {
+    if (!matchingFossilizerClient.name && opts.name) {
+      matchingFossilizerClient.name = opts.name;
+    }
+
+    return matchingFossilizerClient.client;
   }
 
-  return {
+  const fossilizerClient = {
     /**
      * Gets information about the fossilizer.
      * @returns {Promise} a promise that resolve with the information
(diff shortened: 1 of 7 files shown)


simonvadee:
I agree that the callbackUrl's configuration should be a bit refactored.

1. I don't think callbackUrl should be required when creating a fossilizer as the agent already calls `fossilizerHttpClient.fossilize()` with a default callbackUrl.

2. Shouldn't the default callback url be specified at the process creation instead of the agent creation ?

2. Currently, if  `options.evidenceCallbackUrl` or `options.agentUrl` are passed at the agent's creation, any process created by this agent will call `fossilizerHttpClient.fossilize()` with either of these variables , hence making `fossilizerHttpClient.opts.callbackUrl` useless.  I'm not sure this is what we want : if we pass a callbackUrl at the fossilizerHttpClient creation, doesn't it mean that we  want to handle its fossils in a custom fashion (eg: save those in an other db) ? In my opinion `fossilizerHttpClient.opts.callbackUrl` should override the other ones.

t-bast:
Thanks for the comments !

I agree with you, it looks like we currently provide 3 levels where we allow an evidence callback url to be specified:
- at the agent level (create with options agentUrl and evidenceCallbackUrl)
- at the process level (Process constructor with options agentUrl and evidenceCallbackUrl) : if none provided it will use the agent's values
- at the fossilizer http client level (options callbackUrl)

I think this makes sense to allow different levels of configurability and we should use:
- the fossilizer callbackUrl if provided, otherwise
- the process evidenceCallbackUrl if provided, otherwise
- the agent evidenceCallbackUrl if provided, otherwise
- the agent agentUrl (should always be provided right?)

The issue you mention in 3) is that there is another place where we can pass an url: directly as argument to the fossilizerHttpClient's fossilize function, and processes usually provide their process.evidenceCallbackUrl here making the fossilizer's evidenceCallbackUrl useless.

If I understand correctly, I don't think the fossilizer's callback url should be able to change once the fossilizer http client is created. So I would suggest to remove the callbackUrl argument to its fossilize method and only allow setting that value once, at creation time.
Does that make sense?

such:
I think it makes sense.
Just fyi, the target is to get rid of the fossilizer callback url: https://zube.io/stratumn/indigo-framework/c/29

packages/agent-js/src/httpServer.js
```diff
@@ -73,27 +85,61 @@ export default function httpServer(agent, opts = {}) {
...
+          .filter(f => f !== null);
+      }
+
+      return fossilizers;
+    };
+
+    app.post('/:process/upload', (req, res) => {
+      const process = validateProcessUpload(req);
+
+      const store = storeHttpClient(req.body.store.url, {
```
such: do we still want to make it possible to reuse an existing store? (and existing storeHttpClient)

t-bast:
Neat! So that refactoring/change will be done later when we remove the callbackUrl entirely.
I can remove the check for an evidenceCallbackUrl in the process upload API then and let it use the agent's default one for now, it will make it easier to make the move to a web socket.
Does that sound good?

packages/agent-js/src/httpServer.js
t-bast: Good point, that should be doable with the work I did previously.
The idea is to have the caller use the 'name' of an existing storeHttpClient, and then I can look it up locally and reuse the object.
Will do that in the next push.

such:
sounds good to me

packages/agent-js/src/httpServer.js
such: If I understand correctly you don't use the name but just the URL?

t-bast: Yes, I was thinking that there is no reason to have multiple store client to the same url (it means duplicates web sockets, memory, etc).
But if you think it's better to have this store re-use be done only when both url and name match I can change that.

such: No that's fine. But the URL param is still mandatory right? You can't pass the name only?

t-bast: Yes it's validated: if missing an http 400 is returned to the client (from the validateProcessUpload function)

t-bast: I removed the "friendly name" feature entirely as it's in fact going to be messy (and it will conflict with the "real" store/fossilizer name that you get when you call getStoreInfo/getFossilizerInfo).
We'll see if we need to add it later, but probably not :)

packages/agent-js/src/storeHttpClient.js
```diff
@@ -31,16 +31,15 @@ import handleResponse from './handleResponse';
...
-  if (!storeHttpClient.availableStores.find(store => store.url === url)) {
-    storeHttpClient.availableStores.push({
-      name: opt.name,
-      url: url
-    });
+export default function storeHttpClient(url) {
+  // If we already have an existing store for that url, re-use it
+  const matchingStoreClient = storeHttpClient.availableStores.find(
+    store => store.url === url
+  );
```
such: same here?

packages/agent-js/src/fossilizerHttpClient.js
```diff
@@ -81,6 +77,13 @@ export default function fossilizerHttpClient(url, opts = {}) {
       });
     }
   };
+
+  fossilizerHttpClient.availableFossilizers.push({
+    url: url,
+    client: fossilizerClient
+  });
```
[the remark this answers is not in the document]
t-bast: Good point !

packages/agent-js/src/httpServer.js
```diff
@@ -73,27 +85,54 @@ export default function httpServer(agent, opts = {}) {
...
       try {
-        const processModule = new module.constructor();
-        /* eslint-disable no-underscore-dangle */
-        processModule._compile(
-          Buffer.from(req.body.script, 'base64').toString(),
-          '' // this parameter is actually required (undefined isn't accepted)
-        );
-        /* eslint-enable no-underscore-dangle */
-        exported = processModule.exports;
+        process = parseProcessScript(req.body.script);
```
such: sorry about the late realization but I think this should be named `actions`. Does it make sense?

t-bast: Yep no problems, this will be more consistent with what we already have. Updating this.

packages/agent-js/src/httpServer.js
```diff
@@ -64,36 +65,74 @@ export default function httpServer(agent, opts = {}) {
...
+            return null;
+          })
+          .filter(f => f !== null);
+      }
+
+      return fossilizers;
+    };
+
+    app.post('/:process/upload', (req, res) => {
+      const process = validateProcessUpload(req);
```
such: I think this one should also be called actions

t-bast:
Thanks for the reviews ! Let's do the plugins part now and we have a feature ;)

PR #75, opened 2017-10-25
agent-ui: agent info page

t-bast:
* Removed all the "old" components and tests from the old implementation
* Added some content and logic to the agent info page (display agent info and allow refresh)


packages/agent-ui/src/containers/agentInfoPage.js
```diff
@@ -3,32 +3,52 @@ import PropTypes from 'prop-types';
...
 import { withRouter, NavLink } from 'react-router-dom';
 
-export const AgentInfoPage = ({ name, url }) => (
+import { getAgent } from '../actions';
+
+export const AgentInfoPage = ({ name, url, status, fetchAgent }) => (
   <div>
     <p>{name}</p>
     <p>{url}</p>
+    {status !== 'LOADED' && (
```
jeremie-stratumn: maybe also handle case where it is loading .. ?

packages/agent-ui/src/containers/agentInfoPage.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+import { AgentInfoPage } from './agentInfoPage';
+
+chai.use(sinonChai);
+
+describe('<AgentInfoPage />', () => {
+  it('displays the agent name and url', () => {
+    const agentName = 'test';
+    const agentUrl = 'http://localhost:3000';
+    const agentInfoPage = mount(
```
jeremie-stratumn: general question: why not use shallow rendering here? http://airbnb.io/enzyme/docs/api/shallow.html

packages/agent-ui/src/containers/agentInfoPage.js
t-bast: Good idea, let's do it in this PR

packages/agent-ui/src/containers/agentInfoPage.test.js
t-bast: Because shallow rendering won't render my children (ie won't render the paragraph with the name and url). Or does it? I'll double-check :)

t-bast: Cool you were totally right, shallow will simply not render child *react* components but will render normal html components. So I can change almost all my tests to use shallow rendering instead.
I'll send a separate PR for the existing tests that can use shallow instead of mount, thanks for pointing it out.

PR #80, opened 2017-10-25
agent-ui: add process segments page

jeremie-stratumn:
very similar to https://github.com/stratumn/indigo-js/pull/77 for process segments page

### DIFF
--- a/packages/agent-ui/src/actions/getSegments.test.js
+++ b/packages/agent-ui/src/actions/getSegments.test.js
@@ -30,7 +30,6 @@ describe('getSegments action', () => {
 
     return getSegments('foo')(dispatchSpy, getStateStub).then(() => {
       expect(stratumnClientStub.callCount).to.equal(1);
-      expect(getStateStub.callCount).to.equal(1);
     });
   });
 
@@ -65,7 +64,7 @@ describe('getSegments action', () => {
     });
   });
 
-  it('dispatches a failure action when fetching maps fails', () => {
+  it('dispatches a failure action when fetching segments fails', () => {
     const findSegmentsStub = sinon.stub().rejects('Unreachable');
     stratumnClientStub.resolves({
       getProcess: () => ({
@@ -82,7 +81,7 @@ describe('getSegments action', () => {
     });
   });
 
-  it('dispatches a success action when fetching maps succeeds', () => {
+  it('dispatches a success action when fetching segments succeeds', () => {
     const findSegmentsStub = sinon.stub().resolves(['1', '2']);
     stratumnClientStub.resolves({
       getProcess: () => ({
--- a/packages/agent-ui/src/actions/getSegments.test.js
+++ b/packages/agent-ui/src/actions/getSegments.test.js
@@ -75,7 +75,7 @@ describe('getSegments action', () => {
     getStateStub.returns({ agents: { foo: { url: '' } } });
 
     return getSegments('foo', 'bar')(dispatchSpy, getStateStub).then(() => {
-      expect(findSegmentsStub.calledOnce).to.be.true;
+      expect(findSegmentsStub.callCount).to.equal(1);
       expect(dispatchSpy.callCount).to.equal(2);
       const lastActionDispatched = dispatchSpy.getCall(1).args[0];
       expect(lastActionDispatched.type).to.equal(actionTypes.SEGMENTS_FAILURE);
@@ -95,7 +95,7 @@ describe('getSegments action', () => {
       dispatchSpy,
       getStateStub
     ).then(() => {
-      expect(findSegmentsStub.calledOnce).to.be.true;
+      expect(findSegmentsStub.callCount).to.equal(1);
       expect(dispatchSpy.callCount).to.equal(2);
       const lastActionDispatched = dispatchSpy.getCall(1).args[0];
       expect(lastActionDispatched.type).to.equal(actionTypes.SEGMENTS_SUCCESS);
--- a/packages/agent-ui/src/actions/getSegments.test.js
+++ b/packages/agent-ui/src/actions/getSegments.test.js
@@ -1,0 +1,105 @@
+import * as StratumnAgentClient from 'stratumn-agent-client';
+
+import chai, { expect } from 'chai';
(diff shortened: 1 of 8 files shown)


packages/agent-ui/src/actions/getSegments.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+    return getSegments('foo')(dispatchSpy, getStateStub).then(() => {
+      expect(dispatchSpy.callCount).to.equal(2);
+
+      const lastActionDispatched = dispatchSpy.getCall(1).args[0];
+      expect(lastActionDispatched.type).to.equal(actionTypes.SEGMENTS_FAILURE);
+    });
+  });
+
+  it('calls getState', () => {
```
t-bast: You are already testing this in your first unit test. Either remove line 33 and keep this test or remove this test.

packages/agent-ui/src/actions/getSegments.test.js
```diff
@@ -0,0 +1,105 @@
...
+    stratumnClientStub.rejects('Unreachable');
+    getStateStub.returns({ agents: { foo: { url: '' } } });
+
+    return getSegments('foo')(dispatchSpy, getStateStub).then(() => {
+      expect(getStateStub.callCount).to.equal(1);
+    });
+  });
+
+  it('dispatches a failure action when fetching maps fails', () => {
```
t-bast: nit: you meant segments, not map :)
Same below

packages/agent-ui/src/containers/processSegmentsPage.js
```diff
@@ -1,13 +1,56 @@
...
+import { statusTypes } from '../reducers';
+
+export class ProcessSegmentsPage extends Component {
+  componentDidMount() {
+    const { agent, process, fetchSegments } = this.props;
+    fetchSegments(agent, process);
+  }
+
+  render() {
+    const { segments: { status, segments: details, error } } = this.props;
```
t-bast: I think it would make more sense to put the error field in the top level "segments" property (I'm not a big fan of the "segments" property inside a property already named "segments").

t-bast: In fact maybe I'm not understanding this syntax correctly. For me this line means the props object will look like:
{
  segments: {
    status: '',
    segments: {
      details: {},
      error: {}
    }
  }
}
But the propTypes below doesn't show this structure...

packages/agent-ui/src/reducers/segments.test.js
```diff
@@ -0,0 +1,51 @@
...
+import { statusTypes } from './';
+
+describe('segments reducer', () => {
+  it('returns previous state for unknown action', () => {
+    const initialState = { random: 'junk' };
+    const newState = segments(initialState, { type: 'UNKNOWN_ACTION' });
+    expect(newState).to.eql(initialState);
+  });
+
+  it('update state when fetching maps', () => {
```
t-bast: nit: copy/paste: you mean segments, not maps :)

packages/agent-ui/src/reducers/segments.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  it('update state on failure', () => {
+    const initialState = { status: 'foo', segments: [1, 2, 3] };
+    const newState = segments(initialState, {
+      type: actionTypes.SEGMENTS_FAILURE,
+      error: 'unreachable'
+    });
+    const expected = {
+      status: statusTypes.FAILED,
+      segments: [1, 2, 3],
```
t-bast: I think this array should be cleared on errors: it's confusing if the state mentions an error but keeps the successful results of the previous call, don't you think?

packages/agent-ui/src/actions/getSegments.test.js
jeremie-stratumn: will delete l33

packages/agent-ui/src/actions/getSegments.test.js
jeremie-stratumn: yes good catch

packages/agent-ui/src/containers/processSegmentsPage.js
jeremie-stratumn: No this is ES6 nested destructuring. Props looks like:
```
{
  segments: {
    status: '',
    segments: [],
    error: ''
  }
}
```

Since I find it confusing that both the top level structure and the nested structure are named segments, I rename  on the fly the nested `segments` into `details` (this is the `segments: details` syntax). Maybe we can just rename the inner `segments` into `details` at the reducer level. What do you think?

packages/agent-ui/src/reducers/segments.test.js
jeremie-stratumn: yep my bad

packages/agent-ui/src/reducers/segments.test.js
jeremie-stratumn: I agree. The only reason i didn't do it here is because we always dispatch a `REQUEST` action before a `FAILURE` and in the `REQUEST` reducer we set the `segments` to `[]`. So the result on failure will always be `segments: []`. I'm happy to change that behavior to either always set it to `[]` or even delete the field, what do you think?

jeremie-stratumn: As a matter of fact, I think we should always recreate a state from scratch for these reducers, otherwise we may see spurious `error` properties from a previous failure.. For example, imagine, we first request segments and it fails, the state should be:
```
{ ..., segments: { status: 'FAILED', error: 'unreachable' }, ... }
```
Next we retry and it works so we change `status` and add the `details` section, but at no point we remove the error. So it looks like this:
```
{ ..., segments: { status: 'LOADED', error: 'unreachable', details: [...] }, ... }
```
not great right?

packages/agent-ui/src/containers/processSegmentsPage.js
t-bast: Ok, understood. 
I'm good with details, and we can rename that later when we're sure of all the data we'll need in there!

Issue #61, opened 2017-10-18
agent-ui: use redux-logger


jeremie-stratumn:
https://github.com/evgenyrodionov/redux-logger

PR #81, opened 2017-10-25
agent-ui: add redux-logger

jeremie-stratumn:
https://github.com/stratumn/indigo-js/issues/61

### DIFF
--- a/packages/agent-ui/package.json
+++ b/packages/agent-ui/package.json
@@ -13,13 +13,16 @@
     "url": "https://github.com/stratumn/indigo-js/issues"
   },
   "homepage": "https://github.com/stratumn/indigo-js",
-  "keywords": ["stratumn", "agent", "blockchain"],
+  "keywords": [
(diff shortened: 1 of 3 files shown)


t-bast:
I don't think this is necessary. This is already done by the redux devtools extension we're using which has a nicer UI, auto-diff features, etc.

jeremie-stratumn:
the only added value I see is that one needs to install the redux devtools extension to use it on the browser, whereas logger is available to everyone via the console.. I totally agree that the devtools extension is amazing and much better than the logger. but still, could be nice to have the logger around to investigate an issue on someone's machine without having to install the devtools.

t-bast:
I'm honestly not a fan. I don't like things that spam the console, and I don't like feature duplication...installing the extension tools literally takes 10 seconds, so I don't think it's a good reason to use the logger...

jeremie-stratumn:
Fair enough, let's close this then :)

PR #86, opened 2017-10-26
agent-ui: top bar

t-bast:


packages/agent-ui/src/containers/topBar.js
```diff
@@ -1,29 +1,45 @@
...
-      width: 'calc(100% - 240px)',
-      height: '56px',
-      marginLeft: '240px',
-      borderStyle: 'solid'
-    }}
-  >
-    {path}
-  </div>
-);
+const renderTopBarLinks = path => {
```
jeremie-stratumn: We may have to revisit the logic when we add the action button on top right. But that's good for now.

t-bast: Definitely, I haven't thought about it yet but it will be my next task now that the map explorer component is included.

PR #87, opened 2017-10-26
agent-ui: add agents page displays connected agents

t-bast:


packages/agent-ui/src/containers/agentsPage.js
```diff
@@ -42,14 +47,35 @@ export const AgentsPage = ({ fetchAgent }) => {
...
   fetchAgent: PropTypes.func.isRequired
 };
 
-function mapStateToProps(state, ownProps) {
-  console.log('AgentsPage state', state);
-  console.log('AgentsPage ownProps', ownProps);
-  return {};
+export function mapStateToProps(state) {
+  const agents = Object.keys(state.agents || [])
+    .map(a => {
```
jeremie-stratumn: small thing: might be better to start with a filter and then a map. It would make it more easy to understand and not end with `filter(a => a)`. So:
```
.filter(a => state.agents[a].status === LOADED)
.map(a => { name: .. , url: ..} )
```

t-bast: Good point, I'll update that !

packages/agent-ui/src/containers/agentsPage.js
```diff
@@ -62,16 +62,11 @@ AgentsPage.propTypes = {
...
-        };
-      }
-      return null;
-    })
-    .filter(a => a);
+    .filter(a => state.agents[a].status === statusTypes.LOADED)
+    .map(a => ({
+      name: a,
+      url: state.agents[a].url
+    }));
```
jeremie-stratumn: ES6 destructuring:
```
const obj = { foo: { bar: {} }, toto: {} };
const name = 'foo';
const { [name]: {bar} } = obj;
// use bar
...
```

t-bast: Got it. Since here we only want to extract one value (the url) it's simpler to directly access it.
But for future cases we should start using this syntax! 
And maybe javascript will someday support pattern matching ;)

PR #88, opened 2017-10-27
agent-ui: include map explorer

t-bast:
Include the map explorer component inside agent-ui.
I had to fix a very small issue with empty evidences in react mapexplorer.


packages/react-mapexplorer/src/MapExplorer.js
```diff
@@ -150,7 +150,7 @@ class MapExplorer extends Component {
         case 'Evidence':
           segmentContent = (
             <div style={evidencesContent}>
-              {segment.meta.evidences.map(e => renderEvidence(e))}
+              {(segment.meta.evidences || []).map(e => renderEvidence(e))}
```
t-bast: @simonvadee does it make sense that I receive segment.meta without the evidences object (in my test I had no fossilizer)?

PR #105, opened 2017-11-03
agent-ui: fix test warnings

t-bast:
We had a few test warnings, mainly because of import problems: when using an index.js and import from it we're basically importing an object containing all the nested exports, and sometimes it means that some code executes that we don't want to execute in unit tests.
Using file imports in some places fixes these issues.

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/105)

### DIFF
--- a/packages/agent-ui/src/containers/leftNavigation.js
+++ b/packages/agent-ui/src/containers/leftNavigation.js
@@ -50,10 +50,7 @@ LeftNavigation.propTypes = {
   ).isRequired
 };
 
-export function mapStateToProps(state, ownProps) {
-  console.log('LeftNavigation state', state);
-  console.log('LeftNavigation ownProps', ownProps);
-
+export function mapStateToProps(state) {
   let agents = [];
   if (state.agents) {
     agents = Object.keys(state.agents)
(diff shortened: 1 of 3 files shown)


jeremie-stratumn:
:lgtm:
but I think this is going to come back often as someone is going to make the wrong import which will indirectly generate warnings.. I am torn: I like the index.js approach that hides the folder details away but I don't like the console warnings..

Is there a way to treat warnings as errors? Could force the dev to make sure to fix it before merging

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/105#-:-KyK_DF5KYewZnDP9lHl:b-cmrtpp)*

t-bast:
Yeah we'll see if Adrien's presentation today helps us figure out how we want to manage our exports...
Treating warnings as errors can be a good idea too, let's try to find how that would work

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/105#-:-KyKimJ_1garMjwElgVL:b90z8o9)*

PR #139, opened 2017-11-21
agent-js: use Node 8 in Docker images

conord33:
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/139)

### DIFF
--- a/packages/agent-js/Dockerfile.development.alpine
+++ b/packages/agent-js/Dockerfile.development.alpine
@@ -1,4 +1,4 @@
-FROM mhart/alpine-node:6.7.0
+FROM node:8.9.1-alpine
 
 RUN apk add --no-cache git make gcc g++ python
 
--- a/packages/agent-js/Dockerfile.development
+++ b/packages/agent-js/Dockerfile.development
@@ -1,4 +1,4 @@
-FROM node:6.7.0
+FROM node:8.9.1
 
 RUN mkdir /src
 


packages/agent-js/Dockerfile.development.alpine
```diff
@@ -1,4 +1,4 @@
-FROM mhart/alpine-node:6.7.0
```
conord33: not sure if there was a reason to be using this alpine node instead of the official version

t-bast:
Interesting, I don't know indeed if there's a reason for that...you'll probably want to chat with Adrien or Stephan to get a sign-off on that to see if they had a reason to use that alpine image, otherwise :lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/139#-:-KzTwVrEkKaPU0a3Ik34:b38e11a)*

stephan83:
It's because the official image was several hundred megabytes.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/139#-:-KzU-bXIpmL2oTUdvLZN:bi5klxr)*

conord33:
Looks like the alpine differences are a lot smaller now so I think we should stick with the official release if that was the only issue.
```
node 8.9.1-alpine  67.7MB
mhart/alpine-node 8.9.1  66.3MB
```

PR #144, opened 2017-11-28
agent-ui: change display when no maps or segments found

jeremie-stratumn:
Change the display message when no segments or maps are found.

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/144)

### DIFF
--- a/packages/agent-ui/src/components/mapsList.js
+++ b/packages/agent-ui/src/components/mapsList.js
@@ -11,42 +11,47 @@ import Table, {
 } from 'material-ui/Table';
 import Typography from 'material-ui/Typography';
 
-const MapsList = ({ agent, process, mapIds }) => (
-  <Table>
-    <TableHead>
-      <TableRow>
-        <TableCell>
-          <Typography type="subheading">Process maps:</Typography>
-        </TableCell>
-        <TableCell />
-      </TableRow>
-    </TableHead>
-    <TableBody>
-      {mapIds.map(id => (
-        <TableRow key={id} hover>
+const MapsList = ({ agent, process, mapIds }) =>
+  mapIds.length ? (
+    <Table>
+      <TableHead>
+        <TableRow>
           <TableCell>
-            <Typography
-              type="subheading"
-              component={NavLink}
-              to={`/${agent}/${process}/maps/${id}`}
-              style={{ textDecoration: 'none' }}
-            >
-              {id}
-            </Typography>
+            <Typography type="subheading">Process maps:</Typography>
           </TableCell>
-          <TableCell numeric>
-            <Button
-              component={NavLink}
-              to={`/${agent}/${process}/segments?mapIds[]=${id}`}
-            >
-              View segments
-            </Button>
-          </TableCell>
+          <TableCell />
         </TableRow>
-      ))}
-    </TableBody>
-  </Table>
-);
+      </TableHead>
+      <TableBody>
+        {mapIds.map(id => (
+          <TableRow key={id} hover>
+            <TableCell>
+              <Typography
+                type="subheading"
+                component={NavLink}
+                to={`/${agent}/${process}/maps/${id}`}
+                style={{ textDecoration: 'none' }}
+              >
+                {id}
+              </Typography>
+            </TableCell>
+            <TableCell numeric>
+              <Button
+                component={NavLink}
+                to={`/${agent}/${process}/segments?mapIds[]=${id}`}
+              >
+                View segments
+              </Button>
+            </TableCell>
+          </TableRow>
+        ))}
+      </TableBody>
+    </Table>
+  ) : (
+    <Typography type="subheading" style={{ margin: 20 }}>
+      No process map found!
+    </Typography>
+  );
 
 MapsList.defaultProps = {
   mapIds: []
--- a/packages/agent-ui/src/components/mapsList.test.js
+++ b/packages/agent-ui/src/components/mapsList.test.js
@@ -13,10 +13,10 @@ describe('<MapsList />', () => {
     mapIds: ['m1', 'm2']
   };
 
-  const mountMaps = () =>
+  const mountMaps = (props = {}) =>
     mount(
       <MemoryRouter>
-        <MapsList {...requiredProps} />
+        <MapsList {...requiredProps} {...props} />
       </MemoryRouter>
     );
 
@@ -49,5 +49,11 @@ describe('<MapsList />', () => {
 
     verifyLink(0, 'm1');
     verifyLink(1, 'm2');
+  });
+
+  it('does not display map ids when empty', () => {
+    const maps = mountMaps({ mapIds: [] });
+
+    expect(maps.find('Typography')).to.have.length(1);
   });
 });
--- a/packages/agent-ui/src/components/segmentsList.js
+++ b/packages/agent-ui/src/components/segmentsList.js
@@ -29,7 +29,7 @@ export const SegmentsList = ({
         <Typography type="subheading">{`failed to load: ${error}`}</Typography>
       );
     case statusTypes.LOADED:
-      return (
+      return segments.length ? (
         <Table>
           <TableHead>
             <TableRow>
@@ -55,6 +55,10 @@ export const SegmentsList = ({
             ))}
           </TableBody>
         </Table>
+      ) : (
+        <Typography type="subheading" style={{ margin: 20 }}>
+          No process segment found!
+        </Typography>
       );
     default:
       return <CircularProgress className={classes.circular} />;
--- a/packages/agent-ui/src/components/segmentsList.test.js
+++ b/packages/agent-ui/src/components/segmentsList.test.js
@@ -61,4 +61,17 @@ describe('<SegmentsList />', () => {
     verifyLink(0, 's1');
     verifyLink(1, 's2');
   });
+
+  it('does not display segments or filters when empty', () => {
+    const segments = mount(
+      <MemoryRouter>
+        <SegmentsList
+          {...requiredProps}
+          status={statusTypes.LOADED}
+          segments={[]}
+        />
+      </MemoryRouter>
+    );
+    expect(segments.find('Typography')).to.have.length(1);
+  });
 });


piedup:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2HdlkeAe4_Ii38TA7:bb3f1fd) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*
> ```JavaScript
>     </Table>
>   ) : (
>     <Typography type="subheading" style={{ margin: 20 }}>
> ```

I am not an expert but shouldn't style be outside the code?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

piedup:
Reviewed 4 of 4 files at r1.

---

*[packages/agent-ui/src/components/segmentsList.js, line 59 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HpJmQvF3SjZ9zKes:-L-2HpJmQvF3SjZ9zKet:b-ya81ng) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/segmentsList.js#L59)):*
> ```JavaScript
>         </Table>
>       ) : (
>         <Typography type="subheading" style={{ margin: 20 }}>
> ```

Same as above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

t-bast:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---

*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2Hwx6NF9cJ82By3pM:b-o9su04) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*
<details><summary><i>Previously, piedup wrote…</i></summary><blockquote>

I am not an expert but shouldn't style be outside the code?
</blockquote></details>

I think it's good to have it localized like this when it's a small customization

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-:-L-2IKtEsMh3GcnznPaX:bnfp4nl)*

piedup:

---

*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2IuiUir6T1GnBmd3G:br6uqem) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*

I think it's good to have it localized like this when it's a small customization
</blockquote></details>

@t-bast unless it is difficult to have it separated I don't think it is good practice. In this case you'll have to edit mapList.js and segmentList.js to keep design consistency.

How difficult is it to separate from the code?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

bejito:
*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2NOZmcxl6EZYd2GG-:b-ens469) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*
<details><summary><i>Previously, piedup wrote…</i></summary><blockquote>

@t-bast unless it is difficult to have it separated I don't think it is good practice. In this case you'll have to edit mapList.js and segmentList.js to keep design consistency.

How difficult is it to separate from the code?
</blockquote></details>

You should be able to import a stye sheet with a classical `import "xx/yy.css";` and have your style there.  A good practice is to have one stylesheet per component and to put it in the same directory as the component (here it would be `src/components/mapsList.css`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

t-bast:

---

*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2PhAVmDxxPNyKagpC:b-ggq0dy) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*
<details><summary><i>Previously, bejito (Mourad BEJI) wrote…</i></summary><blockquote>

You should be able to import a stye sheet with a classical `import "xx/yy.css";` and have your style there.  A good practice is to have one stylesheet per component and to put it in the same directory as the component (here it would be `src/components/mapsList.css`)
</blockquote></details>

I think this is a different problem. You're mixing the issue of code duplication with the issue of where to set the style.
I think that the style should be set as close to the component as possible (and inline style is ok and even recommended by material UI, but it's a topic where many people disagree so I don't have a strong opinion).
Since Pierre's comment is more that the style is duplicated, I think the right solution would be to extract that small Typography to a dedicated React component that will be used by both the maps and segments pages. This way this component can be changed to be more than a simple Typography if we want, and the change will be reflected in both places.
But I think it's a lot of over-engineer for a margin value :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

bejito:
*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2RVsf3zluA_GOWhmT:byltm8v) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*

I think this is a different problem. You're mixing the issue of code duplication with the issue of where to set the style.
I think that the style should be set as close to the component as possible (and inline style is ok and even recommended by material UI, but it's a topic where many people disagree so I don't have a strong opinion).
Since Pierre's comment is more that the style is duplicated, I think the right solution would be to extract that small Typography to a dedicated React component that will be used by both the maps and segments pages. This way this component can be changed to be more than a simple Typography if we want, and the change will be reflected in both places.
But I think it's a lot of over-engineer for a margin value :)
</blockquote></details>

All right. Usually separating code from style is a good practice, unless your style is dynamic (e.g. dimensions that depend on number of elements) , but that's highly political as you said. 
IMO, having style that is not assigned to classes (or ids) makes the code unclear. If someone wants to change that, it will be harder to find that piece of code in the code base, whereas having all components that need style a class name or an id, and then having style sheets that say how these component should be rendered makes it easier to bulk search (and also avoids duplication).
But again, this is theory.
Anyway, there’s a classic joke that sums up nicely it works in practice :
“Two CSS properties walk into a bar. A barstool in an entirely different bar falls over.”

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

piedup:
*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2Sdvz3zzgf_0doolK:b98e6e2) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*
<details><summary><i>Previously, bejito (Mourad BEJI) wrote…</i></summary><blockquote>

All right. Usually separating code from style is a good practice, unless your style is dynamic (e.g. dimensions that depend on number of elements) , but that's highly political as you said. 
IMO, having style that is not assigned to classes (or ids) makes the code unclear. If someone wants to change that, it will be harder to find that piece of code in the code base, whereas having all components that need style a class name or an id, and then having style sheets that say how these component should be rendered makes it easier to bulk search (and also avoids duplication).
But again, this is theory.
Anyway, there’s a classic joke that sums up nicely it works in practice :
“Two CSS properties walk into a bar. A barstool in an entirely different bar falls over.”

</blockquote></details>

I agree it is a lot of work for juste the margin. This was more of a general comment so that we don't end up with a lot of style within the code at many different places. Let's keep this as is and change it when it becomes material. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

piedup:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-:-L-2ThQwKhcmn6uPogZH:bnfp4nl)*

t-bast:

---

*[packages/agent-ui/src/components/mapsList.js, line 51 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/144#-L-2HdlkeAe4_Ii38TA6:-L-2Ugg2of5SJnX4_ZxY:b-j4zj9m) ([raw file](https://github.com/stratumn/indigo-js/blob/263dc1edde394fb43e54a5aa1076cb5e1ba98529/packages/agent-ui/src/components/mapsList.js#L51)):*
<details><summary><i>Previously, piedup wrote…</i></summary><blockquote>

I agree it is a lot of work for juste the margin. This was more of a general comment so that we don't end up with a lot of style within the code at many different places. Let's keep this as is and change it when it becomes material. 
</blockquote></details>

The idea with agent-ui was to currently have almost no custom style so that we don't have to enter the political debate of how to manage css style in the first place :)
That's why we use Material UI elements, and the only style bits we add are margins/padding.
Our goal is to style only via material UI's template (which is invisible from the component code).
Ideally I agree that even the margins should be factored together, but for now I think it's ok to put some inline since we have almost no custom styling.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/144)*

PR #146, opened 2017-11-29
agent-ui: make it possible to append a new segment/create a new map with Refs

conord33:
Sorry this is so big! In the future I will try to break it up a bit more. This was a good intro into understanding how the frontend is architected though.

This PR adds a ui for add refs when creating maps and segments. This implements that functionality with the combination of a Dialog for actually selecting the segments that will become refs and a smaller component that gives you list of the refs you have selected to add. 

SegmentList was updated to override the default linking behavior. You can now pass it a callback as a prop that will be triggered on the onClick event instead of linking to the segment view.

SegmentFilter was updated to make sure that it can be used in a stateful way and not depend on routing to get the filter options. This prevents the filter options from being cleared when you click on filter while searching for refs.

### Improvements (for future iterations)
* allow refs to be selected from other processes
* provide a way to see more details about the refs you have selected to add and about the segments
  * maybe this can be done in a small view that becomes visible when hovering over certain components
* should we remove the refs field from embedded segments?

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/146)


t-bast:
Great work! Only a few (somewhat minor) comments!

---

Reviewed 30 of 31 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[packages/agent-ui/src/containers/appendSegmentDialog.test.js, line 85 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7F3zHCsoQ7ECtgo3l:-L-7F3zHCsoQ7ECtgo3m:b8bjqd3) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/appendSegmentDialog.test.js#L85)):*
> ```JavaScript
>     const closeDialogSpy = sinon.spy();
>     const dialog = mount(
>       <Provider store={store}>
> ```

It's a pain that we have to add this Provider when it's in fact one of our children (RefChipList) that needs to be connected...
I'm wondering if we should start using shallow rendering and specifically diving in the sub-components we want to render to allow that...it's probably even more of a pain to handle than wrapping in a Provider...

---

*[packages/agent-ui/src/containers/index.js, line 21 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7GAb-_WH5XV423QX3:-L-7GAb-_WH5XV423QX4:b4yymps) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/index.js#L21)):*
> ```JavaScript
> export { default as AppendSegmentDialog } from './appendSegmentDialog';
> export { default as SelectRefsDialog } from './selectRefsDialog';
> export { default as RefChipList } from './refChipList';
> ```

I think that RefChipList should be a component, not a container.
It should receives props via React, not redux.
This would remove the need for the Provider in segment and map list tests.

---

*[packages/agent-ui/src/containers/selectRefsDialog.js, line 113 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7Ixt2V7s3rYGOWp7M:-L-7Ixt2V7s3rYGOWp7N:b-ade1ph) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/selectRefsDialog.js#L113)):*
> ```JavaScript
> }
> 
> export default withRouter(
> ```

Do you really need the withRouter?

---

*[packages/agent-ui/src/reducers/selectRefs.js, line 38 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7Jbu6IZmQZ2RQaBgN:-L-7Jbu6IZmQZ2RQaBgO:b-lyz6fu) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/reducers/selectRefs.js#L38)):*
> ```JavaScript
>     case actionTypes.SELECT_REFS_ADD:
>       return onlyAddNewRef(state, action.ref);
>     case actionTypes.SELECT_REFS_CLEAR:
> ```

maybe you want to return initialState here (to be more future-proof and make sure you don't keep some parts of the state we don't want)?

---

*[packages/agent-ui/src/test/builders/state/testStateBuilder.js, line 19 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7K18GNONroIgLGXim:-L-7K18GNONroIgLGXin:b-bayf6q) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/test/builders/state/testStateBuilder.js#L19)):*
> ```JavaScript
>       request: {}
>     };
>     this.selectRefs = {
> ```

Ideally, it would be cool if you could also add a method to add real test builder support here (instead of doing it custom in tests).
I'd add:

withSelectRefDialog() {
    this.selectRefs.show = true;
    return this;
}

withSelectedRef(ref) {
    this.selectRefs.refs.push(ref);
    return this;
}

---

*[packages/agent-ui/src/utils/queryString.js, line 8 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7KciKpTbsb-pJoqzC:-L-7KciKpTbsb-pJoqzD:b-yvx3ik) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/utils/queryString.js#L8)):*
> ```JavaScript
> export const parse = o => qs.parse(o, { ignoreQueryPrefix: true });
> 
> export const parseAgentAndProcess = uri => {
> ```

That's useful, thanks :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-:-L-7Kf-kQUemVY7pARwP:b-ur5dws)*

conord33:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[packages/agent-ui/src/containers/appendSegmentDialog.test.js, line 85 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7F3zHCsoQ7ECtgo3l:-L-7MxXiDpgDdEfRoUUS:b-uuf5n1) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/appendSegmentDialog.test.js#L85)):*

It's a pain that we have to add this Provider when it's in fact one of our children (RefChipList) that needs to be connected...
I'm wondering if we should start using shallow rendering and specifically diving in the sub-components we want to render to allow that...it's probably even more of a pain to handle than wrapping in a Provider...
</blockquote></details>

i think i have a good solution for this for mocking it.

---

*[packages/agent-ui/src/containers/index.js, line 21 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7GAb-_WH5XV423QX3:-L-7Negn5mm48LScwL1c:b-2zq1lz) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/index.js#L21)):*

I think that RefChipList should be a component, not a container.
It should receives props via React, not redux.
This would remove the need for the Provider in segment and map list tests.
</blockquote></details>

i was going back and forth on this a bunch, but deciding to take this route to keep it more self contained. if it doesnt use redux then we will need to repeat essentially the same logic in all three of the dialogs and always pass RefChipList the same thing.

---

*[packages/agent-ui/src/containers/selectRefsDialog.js, line 113 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7Ixt2V7s3rYGOWp7M:-L-7NTGryyT3cSYmndJK:by7p89s) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/selectRefsDialog.js#L113)):*

Do you really need the withRouter?
</blockquote></details>

i wasnt sure about that. right now I am only using it to get the current process and agent. is there a better mechanism to do that?

---

*[packages/agent-ui/src/reducers/selectRefs.js, line 38 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7Jbu6IZmQZ2RQaBgN:-L-7NE4Atf-skGQ0YGvq:b-569fpf) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/reducers/selectRefs.js#L38)):*

maybe you want to return initialState here (to be more future-proof and make sure you don't keep some parts of the state we don't want)?
</blockquote></details>

the initial state also contains the show option. so in this case I would also close the dialog

---

*[packages/agent-ui/src/test/builders/state/testStateBuilder.js, line 19 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7K18GNONroIgLGXim:-L-7ONra4pmEYD6SUhJu:b-d7bhn0) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/test/builders/state/testStateBuilder.js#L19)):*

Ideally, it would be cool if you could also add a method to add real test builder support here (instead of doing it custom in tests).
I'd add:

withSelectRefDialog() {
    this.selectRefs.show = true;
    return this;
}

withSelectedRef(ref) {
    this.selectRefs.refs.push(ref);
    return this;
}
</blockquote></details>

will do

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146)*

t-bast:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[packages/agent-ui/src/containers/index.js, line 21 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7GAb-_WH5XV423QX3:-L-7T8chfFUMKpxYC5d-:bq80idl) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/index.js#L21)):*
<details><summary><i>Previously, conord33 (Conor Dawson) wrote…</i></summary><blockquote>

i was going back and forth on this a bunch, but deciding to take this route to keep it more self contained. if it doesnt use redux then we will need to repeat essentially the same logic in all three of the dialogs and always pass RefChipList the same thing.
</blockquote></details>

Yeah it's true...I'm fine with keeping it a container then

---

*[packages/agent-ui/src/containers/selectRefsDialog.js, line 113 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7Ixt2V7s3rYGOWp7M:-L-7TEv_EntBdGhLL85i:b-6j8jis) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/containers/selectRefsDialog.js#L113)):*
<details><summary><i>Previously, conord33 (Conor Dawson) wrote…</i></summary><blockquote>

i wasnt sure about that. right now I am only using it to get the current process and agent. is there a better mechanism to do that?
</blockquote></details>

If you're only getting the path, you don't need withRouter for that (since you're parsing the path manually to get agent and process)

---

*[packages/agent-ui/src/reducers/selectRefs.js, line 38 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7Jbu6IZmQZ2RQaBgN:-L-7TPGXzN-1F7i_GXPH:b-21vs15) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/reducers/selectRefs.js#L38)):*
<details><summary><i>Previously, conord33 (Conor Dawson) wrote…</i></summary><blockquote>

the initial state also contains the show option. so in this case I would also close the dialog
</blockquote></details>

Good point, clear doesn't always also close. Ok to keep it this way then!

---

*[packages/agent-ui/src/test/builders/state/testStateBuilder.js, line 19 at r2](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7K18GNONroIgLGXim:-L-7TVakujNxScjRfgXM:b2nhpa7) ([raw file](https://github.com/stratumn/indigo-js/blob/107fd89189b68d13840cab1d1c0c99c0e98e817f/packages/agent-ui/src/test/builders/state/testStateBuilder.js#L19)):*
<details><summary><i>Previously, conord33 (Conor Dawson) wrote…</i></summary><blockquote>

will do
</blockquote></details>

Thx

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146)*

conord33:
Review status: 29 of 32 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[packages/agent-ui/src/test/stubComponent.js, line 1 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7VRrWVI_U4QloXsqy:-L-7VRrWVI_U4QloXsqz:b-44mjtd) ([raw file](https://github.com/stratumn/indigo-js/blob/b17cb1c9e4934518c362246d5af88be94fb302ca/packages/agent-ui/src/test/stubComponent.js#L1)):*
> ```JavaScript
> import sinon from 'sinon';
> ```

@t-bast we can use this to stub out full containers to avoid wrapping things in providers just to mount components

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146)*

jeremie-stratumn:
Review status: 0 of 32 files reviewed at latest revision, 10 unresolved discussions.

---

*[packages/agent-ui/src/components/segmentsFilter.js, line 39 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7aA2c1VKKjzFKK7qG:-L-7aA2c1VKKjzFKK7qH:b442v8p) ([raw file](https://github.com/stratumn/indigo-js/blob/b17cb1c9e4934518c362246d5af88be94fb302ca/packages/agent-ui/src/components/segmentsFilter.js#L39)):*
> ```JavaScript
>   componentWillReceiveProps(nextProps) {
>     const { filters } = nextProps;
>     // Only overwrite our state if we are passed filters as a prop
> ```

Why do you have to do that?

---

*[packages/agent-ui/src/components/segmentsList.js, line 89 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7bPkaVYaSCGPy9wkp:-L-7bPkaVYaSCGPy9wkq:b-wcqzvz) ([raw file](https://github.com/stratumn/indigo-js/blob/b17cb1c9e4934518c362246d5af88be94fb302ca/packages/agent-ui/src/components/segmentsList.js#L89)):*
> ```JavaScript
>   error: PropTypes.string,
>   handleClick: PropTypes.func,
>   segments: PropTypes.arrayOf(
> ```

why do you have to change segments into a shape? It seems you always only need the `meta.linkHash`.

---

*[packages/agent-ui/src/reducers/segments.js, line 19 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7dhSmztkXCj3_-eoJ:-L-7dhSmztkXCj3_-eoK:b-wx56e0) ([raw file](https://github.com/stratumn/indigo-js/blob/9d76e6352e2104a082212fc5c7d242b6cb432e08/packages/agent-ui/src/reducers/segments.js#L19)):*
> ```JavaScript
>       return {
>         status: statusTypes.LOADED,
>         details: action.segments.map(({ meta: { linkHash } }) => linkHash)
> ```

Not sure what you are gaining by keeping the whole segments..?

---

*[packages/agent-ui/src/utils/queryString.js, line 9 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7f4xBjc2c9rhhRZZ8:-L-7f4xBjc2c9rhhRZZ9:b1a93eh) ([raw file](https://github.com/stratumn/indigo-js/blob/b17cb1c9e4934518c362246d5af88be94fb302ca/packages/agent-ui/src/utils/queryString.js#L9)):*
> ```JavaScript
> 
> export const parseAgentAndProcess = uri => {
>   const match = matchPath(uri, { path: '/:agent/:process' });
> ```

Do you know what happens if I do `parseAgentAndProcess('/foo/bar/hey')`?
Will `process` be `bar` or `foo/bar`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146)*

t-bast:
:lgtm:, waiting for Jérémie's comments

---

Reviewed 6 of 6 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-:-L-7ht_iq6-l1vJno5VU:bvy03k4)*

conord33:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[packages/agent-ui/src/components/segmentsList.js, line 89 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7bPkaVYaSCGPy9wkp:-L-7jG8p5BMnm2pKoxsA:b-c3coxl) ([raw file](https://github.com/stratumn/indigo-js/blob/b17cb1c9e4934518c362246d5af88be94fb302ca/packages/agent-ui/src/components/segmentsList.js#L89)):*
<details><summary><i>Previously, jeremie-stratumn (Jeremie Dayan) wrote…</i></summary><blockquote>

why do you have to change segments into a shape? It seems you always only need the `meta.linkHash`.
</blockquote></details>

it is the only required property. the rest of the segment is just passed through to the handleClick prop if it was passed in.

---

*[packages/agent-ui/src/reducers/segments.js, line 19 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7dhSmztkXCj3_-eoJ:-L-7iQjM7cMmXwVtzON3:b-wc6rp8) ([raw file](https://github.com/stratumn/indigo-js/blob/9d76e6352e2104a082212fc5c7d242b6cb432e08/packages/agent-ui/src/reducers/segments.js#L19)):*
<details><summary><i>Previously, jeremie-stratumn (Jeremie Dayan) wrote…</i></summary><blockquote>

Not sure what you are gaining by keeping the whole segments..?
</blockquote></details>

the whole segment is needed so you can pass the segment to the handleClick method in the segmentList. this allows you to do more complicated stuff like adding segments as refs. the alternative would be to make an api call to get each segment you have but seems kinda pointless if we already have all the segments at one point

---

*[packages/agent-ui/src/utils/queryString.js, line 9 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7f4xBjc2c9rhhRZZ8:-L-7izwcsNueO2sqAVb3:b59p21q) ([raw file](https://github.com/stratumn/indigo-js/blob/b17cb1c9e4934518c362246d5af88be94fb302ca/packages/agent-ui/src/utils/queryString.js#L9)):*
<details><summary><i>Previously, jeremie-stratumn (Jeremie Dayan) wrote…</i></summary><blockquote>

Do you know what happens if I do `parseAgentAndProcess('/foo/bar/hey')`?
Will `process` be `bar` or `foo/bar`?
</blockquote></details>

i think it will still work but ill add some tests

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146)*

jeremie-stratumn:
:lgtm:

---

Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[packages/agent-ui/src/reducers/segments.js, line 19 at r3](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-L-7dhSmztkXCj3_-eoJ:-L-7lE9b0jbEUVOoAeBH:b-6o8fvb) ([raw file](https://github.com/stratumn/indigo-js/blob/9d76e6352e2104a082212fc5c7d242b6cb432e08/packages/agent-ui/src/reducers/segments.js#L19)):*
<details><summary><i>Previously, conord33 (Conor Dawson) wrote…</i></summary><blockquote>

the whole segment is needed so you can pass the segment to the handleClick method in the segmentList. this allows you to do more complicated stuff like adding segments as refs. the alternative would be to make an api call to get each segment you have but seems kinda pointless if we already have all the segments at one point
</blockquote></details>

true. makes sense. Thank you

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-:-L-7lPLHbLe1bCcZGVp9:bnfp4nl)*

t-bast:
:lgtm:

---

Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/146#-:-L-AuUDXp6rVwWPC32qN:bnfp4nl)*

PR #163, opened 2017-12-05
agent-ui: Fix selectRefsDialog for undef agent

bejito:
In some cases, the agent can not be initialized in the store

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/163)

### DIFF
--- a/packages/agent-ui/src/containers/selectRefsDialog.js
+++ b/packages/agent-ui/src/containers/selectRefsDialog.js
@@ -115,9 +115,9 @@ SelectRefsDialog.propTypes = {
 function mapStateToProps(state, ownProps) {
   const { location: { pathname } } = ownProps;
   const { process, agent } = parseAgentAndProcess(pathname);
-  const { segments, selectRefs: { show } } = state;
-  const { agents: { [agent]: { processes } } } = state;
-  return { agent, process, segments, show, processes: Object.keys(processes) };
+  const { segments, selectRefs: { show }, agents } = state;
+  const processes = agents[agent] ? Object.keys(agents[agent].processes) : [];
+  return { agent, process, segments, show, processes };
 }
 
 export default connect(mapStateToProps, {


conord33:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/163#-:-L-bpDFs-Ob93W0k2enZ:bnfp4nl)*

t-bast:
Can you detail in what scenario that happens?
Loading the agents is supposed to be the first thing the main component does, so I'm a bit surprised...unless selectRefsDialog renders very early (probably sooner than it should)

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/163#-:-L-erhQ7JyqD7O_zLZMe:b-ba4d4j)*

packages/agent-ui/src/containers/selectRefsDialog.js
```diff
@@ -115,9 +115,9 @@ SelectRefsDialog.propTypes = {
 function mapStateToProps(state, ownProps) {
   const { location: { pathname } } = ownProps;
   const { process, agent } = parseAgentAndProcess(pathname);
-  const { segments, selectRefs: { show } } = state;
-  const { agents: { [agent]: { processes } } } = state;
```
conord33: @t-bast it happens here when you load the page. because `agent` is loaded from the current location (url), when you are at the home page there is no agent present so it returns `undefined`. Then destructuring with undefined throws an error.

t-bast:
But the problem then is that the "agent" variable is undefined, not that the agents in the state are not loaded right?
So the fix should be to check if (agent) instead of agents[agent] ?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/163#-:-L-f-H5bh_XeE1gV0K3E:b-7gt55i)*

bejito:
@t-bast yes, but it is safer this way. It is always better to safe deep get in objects. If we change the way we fill the store one day, this will not break

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/163#-:-L-f0b0OEYHcgTJqt6j8:b-8m2fmw)*

t-bast:
Agreed, it's just that it was hard to understand the problem because the description was misleading :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/163#-:-L-f116xHEJkOxdacJhy:b-mf2on)*

PR #169, opened 2017-12-08
agent-ui: add websocket notifications

jeremie-stratumn:
We open a websocket for each agent in the state. This websocket can dispatch new notification on `didSave` message that will make the `mapexplorer` refresh if someone else added a new segment.

At the moment, 2 notifications are disptached when I add a new segment: 1 from the appendSegment action and 1 from the websocket message we receive on didSave event. It all happens in the same tick so it does not refresh twice but we should fix this IMO.

We could rely on the didSave notification only to refresh the mapexplorer but I don't think it is very reliable, as it can be delayed (if consensus takes long time for ex..). It would be better to have a way to uniquely identify notifications (hash instead of uuid?) and check if we've already seen it.

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/169)

### DIFF
--- a/packages/agent-ui/package.json
+++ b/packages/agent-ui/package.json
@@ -53,6 +53,7 @@
     "enzyme-adapter-react-16": "^1.1.0",
     "eslint": "^4.8.0",
     "eslint-plugin-react": "^7.4.0",
+    "mock-socket": "^7.1.0",
     "react-scripts": "^1.0.17",
     "react-test-renderer": "^16.0.0",
     "redux-mock-store": "^1.3.0",
--- a/packages/agent-ui/package.json
+++ b/packages/agent-ui/package.json
@@ -12,19 +12,17 @@
   "bugs": {
     "url": "https://github.com/stratumn/indigo-js/issues"
   },
-  "keywords": [
-    "stratumn",
-    "agent",
-    "blockchain"
-  ],
+  "keywords": ["stratumn", "agent", "blockchain"],
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --env=jsdom",
-    "test:ci": "react-scripts build && CI=true react-scripts test --env=jsdom --coverage",
+    "test:ci":
+      "react-scripts build && CI=true react-scripts test --env=jsdom --coverage",
     "eject": "react-scripts eject"
   },
   "dependencies": {
+    "babel-polyfill": "^6.26.0",
     "history": "^4.7.2",
     "immutable": "^3.8.2",
     "material-ui": "^1.0.0-beta.21",
@@ -46,7 +44,6 @@
     "uuid": "^3.1.0"
   },
   "devDependencies": {
-    "babel-polyfill": "^6.26.0",
     "chai": "^4.1.2",
     "css.escape": "^1.5.1",
     "enzyme": "^3.1.0",
--- a/packages/agent-ui/src/actions/appendSegment.js
+++ b/packages/agent-ui/src/actions/appendSegment.js
@@ -73,11 +73,7 @@ export const appendSegment = (...args) => (dispatch, getState) => {
         dispatch(appendSegmentSuccess(segment));
         dispatch(closeDialog());
         dispatch(clearRefs());
-        dispatch(
-          addNotifications([
-            makeNewSegmentNotification(agent, process, segment)
-          ])
-        );
+        dispatch(addNotifications([makeNewSegmentNotification(segment)]));
       })
       .catch(err => {
         dispatch(appendSegmentFailure(err));
--- a/packages/agent-ui/src/actions/getAgent.js
+++ b/packages/agent-ui/src/actions/getAgent.js
@@ -1,5 +1,6 @@
 import { getAgent as getAgentClient } from 'stratumn-agent-client';
 import * as actionTypes from '../constants/actionTypes';
+import { openWebSocket, closeWebSocket } from '../utils/webSocketHelpers';
 
 const getAgentRequest = (name, url) => ({
   type: actionTypes.AGENT_INFO_REQUEST,
@@ -19,14 +20,18 @@ const getAgentSuccess = (name, agent) => ({
   agent
 });
 
-export const removeAgent = name => ({
-  type: actionTypes.AGENT_INFO_DELETE,
-  name
-});
+export const removeAgent = name => {
+  closeWebSocket(name);
+  return {
+    type: actionTypes.AGENT_INFO_DELETE,
+    name
+  };
+};
 
 export const getAgent = (name, url) => dispatch => {
   dispatch(getAgentRequest(name, url));
   return getAgentClient(url)
     .then(agent => dispatch(getAgentSuccess(name, agent)))
+    .then(() => openWebSocket(name, url, dispatch))
     .catch(err => dispatch(getAgentFailure(name, err)));
 };
--- a/packages/agent-ui/src/actions/getAgent.test.js
+++ b/packages/agent-ui/src/actions/getAgent.test.js
@@ -6,20 +6,24 @@ import sinonChai from 'sinon-chai';
 
 import { getAgent, removeAgent } from './getAgent';
 import * as actionTypes from '../constants/actionTypes';
+import * as webSocket from '../utils/webSocketHelpers';
 
 chai.use(sinonChai);
 
 describe('getAgent action', () => {
   let stratumnClientStub;
+  let openWsStub;
   let dispatchSpy;
 
   beforeEach(() => {
     stratumnClientStub = sinon.stub(StratumnAgentClient, 'getAgent');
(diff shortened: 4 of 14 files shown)


t-bast:
Reviewed 14 of 14 files at r1.

---

*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L-qd5MiYJ-UcgVisEwP:b1eeuk4) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*
> ```JavaScript
> // re open ws with all exisiting agents in our state.
> // On unmount, we want to close all the opened ws.
> export class WebSocketsManager extends Component {
> ```

Maybe I'm wrong, but I think it's hackish to have a fake React object for that. A React object should always be a UI object in my opinion.
Instead I would recommend doing the socket management in Root or App's componentDidMount / componentWillUnmount methods.
What do you think? Maybe I'm wrong and it's a valid pattern, but it surprises me :)

---

*[packages/agent-ui/src/utils/webSocketHelpers.js, line 22 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qeXOk5pOpsmrGUEVb:-L-qeXOlpyFa7yDFzPbi:b-1toasz) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/utils/webSocketHelpers.js#L22)):*
> ```JavaScript
>       }
>     } catch (e) {
>       // do nothing
> ```

Maybe log to the console? It can be helpful to diagnose prod issues

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169)*

bejito:
*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L-qmWJGH4KSWmWrQ9Bt:bmxmexa) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*

Maybe I'm wrong, but I think it's hackish to have a fake React object for that. A React object should always be a UI object in my opinion.
Instead I would recommend doing the socket management in Root or App's componentDidMount / componentWillUnmount methods.
What do you think? Maybe I'm wrong and it's a valid pattern, but it surprises me :)
</blockquote></details>

You could also make this a mixin and make other root components extend it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169)*

t-bast:

---

*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L-qqO6itgAMpKEGVM6I:b-fqhn4m) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*
<details><summary><i>Previously, bejito (Mourad BEJI) wrote…</i></summary><blockquote>

You could also make this a mixin and make other root components extend it
</blockquote></details>

I don't know what a mixin is but it doesn't look like it's really recommended: https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169)*

bejito:
*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L-qsqA2v6ehPZqpjK7r:b-rz40k0) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*

I don't know what a mixin is but it doesn't look like it's really recommended: https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750 
</blockquote></details>

yeah, just an abstract class is actually good enough :)
I mean extend `React.Component` with this code, e.g. call it `ConnectedComponent`
And then, when you need a component that is connected to the ws, probably the root of the view as you said, you make it extend `ConnectedComponent`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169)*

jeremie-stratumn:

---

*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L-rGAI5SKdxZfnAcEjo:b-jjc077) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*
<details><summary><i>Previously, bejito (Mourad BEJI) wrote…</i></summary><blockquote>

yeah, just an abstract class is actually good enough :)
I mean extend `React.Component` with this code, e.g. call it `ConnectedComponent`
And then, when you need a component that is connected to the ws, probably the root of the view as you said, you make it extend `ConnectedComponent`
</blockquote></details>

I think I see what you mean @bejito. The thing is,  WebSocketsManager is a connected component and uses a mapStateToProps to extract agent info from state. I cannot seem to extend a connected component, I always get a runtime error. WebSocketsManager could just be an unconnected component but I would need to pass info from props and both root nor app are connected component either..

@t-bast option would hit the same problem: root / app are not connected so I cannot access agent info.. Do I really want to make them connected for that purpose?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169)*

jeremie-stratumn:

---

*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L-rIryB8f6o24DVE5Wp:b6vco7b) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*
<details><summary><i>Previously, jeremie-stratumn (Jeremie Dayan) wrote…</i></summary><blockquote>

I think I see what you mean @bejito. The thing is,  WebSocketsManager is a connected component and uses a mapStateToProps to extract agent info from state. I cannot seem to extend a connected component, I always get a runtime error. WebSocketsManager could just be an unconnected component but I would need to pass info from props and both root nor app are connected component either..

@t-bast option would hit the same problem: root / app are not connected so I cannot access agent info.. Do I really want to make them connected for that purpose?
</blockquote></details>

My bad, there is something weird when I `import { WebSocketsManager } from './'`. Instead if I `import WebSocketsManager from './.'` it does not complain. The thing is, now app extends WebSocketsManager but  WebSocketsManager.componentDidMount() is never called..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169)*

t-bast:
:lgtm:

---

---

*[packages/agent-ui/src/containers/webSocketsManager.js, line 13 at r1](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-L-qd5Mg8q24ywsr1XqP:-L03bQsjL24YWAQLRGkH:b4oqjf1) ([raw file](https://github.com/stratumn/indigo-js/blob/13a7841c7970a02060cb07272367d7002b88e8cc/packages/agent-ui/src/containers/webSocketsManager.js#L13)):*
<details><summary><i>Previously, jeremie-stratumn (Jeremie Dayan) wrote…</i></summary><blockquote>

My bad, there is something weird when I `import { WebSocketsManager } from './'`. Instead if I `import WebSocketsManager from './.'` it does not complain. The thing is, now app extends WebSocketsManager but  WebSocketsManager.componentDidMount() is never called..
</blockquote></details>

I see what you mean, it's probably not worth making app or root a connected component for that.
I'm good with your current solution (I don't think we should go the mixin or abstract class way as it's not recommended).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-:-L03baDw5bTqjoPutPOw:bnfp4nl)*

bejito:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/169#-:-L03lEZJ4Clp6J9KRT26:bnfp4nl)*

PR #190, opened 2017-12-20
release: update package.json for publish

t-bast:
Scoped packages cannot be private if we want lerna to publish them.
Scoped packages need the publishConfig in each package.json according to lerna's documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/190)

### DIFF
--- a/packages/agent-ui/package.json
+++ b/packages/agent-ui/package.json
@@ -23,9 +23,7 @@
     "test:ci": "react-scripts build && CI=true react-scripts test --env=jsdom --coverage",
     "eject": "react-scripts eject"
   },
-  "publishConfig": {
-    "access": "public"
-  },
+  "private": true,
   "dependencies": {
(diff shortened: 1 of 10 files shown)


packages/agent-ui/package.json
```diff
@@ -4,7 +4,6 @@
   "description": "Stratumn agent user interface",
   "author": "Stratumn team",
   "license": "Apache-2.0",
-  "private": true,
```
such: we don't really need to publish this package in theory since we provide the docker image

t-bast:
That's a good point. I'll remove it.

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/190#-:-L0oUQu7g9Y8NUa1mOdi:bp4gy9v)*

PR #198, opened 2017-12-21
agent-ui: use alpine to reduce docker image size

such:
Uncompressed to size goes from ~900MB to ~500MB

No additional apk package seems to be necesssary. Only the `fsevents` dependency fails to be installed but I don't think it's a big deal inside that Docker image.

If we wanted to install it we'd have to add:
`RUN apk add --no-cache git make gcc g++`

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/198)

### DIFF
--- a/packages/agent-ui/Dockerfile
+++ b/packages/agent-ui/Dockerfile
@@ -1,6 +1,4 @@
-FROM node:8
-
-RUN npm install -g yarn
+FROM node:8-alpine
 
 RUN mkdir /agent-ui
 ADD package.json /agent-ui/package.json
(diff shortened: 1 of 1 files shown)


t-bast:
fsevent isn't an issue, it fails locally on my linux machine as well because it's something that's simply not needed on linux
:lgtm: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/198#-:-L0thK3se5a9o387ZpUP:b-wbdtac)*

PR #216, opened 2017-12-28
agent-ui: properly handle refresh on all routes

### LINKED ISSUE #214: agent-ui: refresh segments page and left drawer contents disappears
I noticed that when refreshing the segments or maps page the left drawer has no content anymore. This is due to https://github.com/stratumn/indigo-js/pull/182 I believe. The AgentsPage is in charge of fetching STALE agents but it is not always mounted, for example when refreshing the segments page it is not.. 

Should create a class `AgentsInitializer` that always renders (not behind a route) null and fetch stale agents on mount.

Now I can go back to my vacations :)

t-bast:
The main issue was that only the agents page (/) handled loading the agents.
It should be the responsibility of the web sockets manager : 
* When the app mounts, all agents have been loaded from the persisted state and are STALE
* So we should fetch each one of them, which will open a web socket as well
* At the end of the app's life, we should close those web sockets

Resolves #214 

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/216)

### DIFF
--- a/packages/agent-ui/src/containers/agentsPage.js
+++ b/packages/agent-ui/src/containers/agentsPage.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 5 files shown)


such:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/216#-:-L1SwDlE5i01FNls2bdX:bnfp4nl)*

PR #218, opened 2017-12-28
agent-ui: error dialog appending without selection

### LINKED ISSUE #157: agent-ui: wrong linkHash on append when changing process
### Expected behavior
When I am on a map, I should be able to append a new segment

### Actual behavior
When updating a map, and then switching directly to another map from another process, we can use the append button even if no segment is selected. In this case, the store still has the previous segment's linkHash, and tries to look for it in the current process, which results in a 404 (/{process2}/segments/{hash from a segment from process 1})

t-bast:
We never dealt properly with an attempt to append without selecting a segment or with a last selected segment from another process.
We are now displaying an error message explaining that a parent segment needs to be selected.
The error message dialog can probably be refactored to a common component later if we need it elsewhere (which will eventually happen).

Resolves #157 

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/indigo-js/218)


such:
:lgtm:

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/stratumn/indigo-js/218#-:-L1TR4R2kA-b3Ky-bu-Y:bnfp4nl)*

PR #257, opened 2018-03-19
map-explorer: more data in TM evidence

t-bast:
Adding more data in TM evidence. The most useful one is probably the validators/expected validators section that can help quickly see that a validator is missing.
If we want the evidence to be easily understood by anyone, I think we need to do quite a bunch of design work to figure out how to visualize it. It will probably involve an interactive component to let the user dive into evidence parts, so it's quite a bit of work that requires better design skills than mine :)
That's also something that wouldn't be specific to us but would work for every project that uses Tendermint, so it might be worth doing that in a completely separate repo?

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/js-indigocore/257)

### DIFF
--- a/packages/agent-ui/yarn.lock
+++ b/packages/agent-ui/yarn.lock
@@ -2,34 +2,6 @@
 # yarn lockfile v1
 
 
-"@indigoframework/client@^0.2.0":
-  version "0.2.0"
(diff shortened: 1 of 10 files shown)


such:
:lgtm:

---

Reviewed 10 of 10 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/stratumn/js-indigocore/257#-:-L7yHw0n2IT3MXI6w_6p:bnfp4nl)*

PR #267, opened 2018-04-03
agent-js: this.append(type) and plugin update

simonvadee:
1. the `append(state, meta, type)` available in the agent's actions was changed to `append(type)`: the first two parameters were almost never used and it didn't make sense to do `this.append(null, null, newType)` to update the segment's type.

2. plugin `signedState` was removed

3. other plugins were updated so that they follow the new link.meta structure

---
This change is [Reviewable](https://reviewable.io/reviews/stratumn/js-indigocore/267)


such:
:lgtm:

---

Reviewed 17 of 17 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/stratumn/js-indigocore/267#-:-L9BAgHC4phFfR1Kc_-4:bnfp4nl)*

